PR #2, opened 2020-03-20
Use GitHub workflows as CI

anders-kiaer:
Removes the previous Travis CI workflow, and converts it into GitHub actions.

Caching is now simplified (no more need for calculating `md5sum setup.py/build_environment.sh` manually).

Note that when re-running a PR (i.e. adding new commits), caching might/will currently silently fail due to https://github.com/actions/cache/issues/208 (i.e. compilations are done again).


.github/workflows/flownet.yml
```diff
@@ -0,0 +1,77 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - master
+  
+jobs:
+  FlowNet:
+    runs-on: ubuntu-latest
```
wouterjdb: Maybe fix on 18.10?

PR #6, opened 2020-03-24
Add individual EQUIL support

wouterjdb:
This PR adds support for individual tuning per tube of the equilibrium and is aimed at testing whether this makes it easier to history match the Norne case that is attached in the FlowNet repository.

- [x] Add support for individual EQUIL tuning
- [x] Add THPRES keyword to make equilibration stable
- [x] Find solution for repeating of RSVD keyword


wouterjdb:
@anders-kiaer When you have multiple EQLNUM regions, you will have to specify RSVD tables for each EQLNUM. This RSVD table is a (single) static table currently. Any preferences on how we repeat that table?

Possible solution could be to:

- check whether there is only a single entry, if so, then we repeat that entry _len(EQLNUM)_ times. 

- When there are multiple entries we assume that the user knows what he/she is doing and we just copy-paste the file (which might of course then lead to a mismatch in the number of tables).

anders-kiaer:
> Any preferences on how we repeat that table?

Good question. I'm thinking: maybe this could be the first step towards PVT from non-Eclipse sources? E.g.

1. the user provides a "RSVD table" in the form of e.g. a `csv` file, which is then pointed to through the configuration file.
2. The `EQUIL` parameter gets this `RSVD` data, and creates + write out a suitable `RSVD` table (using the standard framework for going from parameters to rendered simulation input).

Then problem is solved, and we are at the same time one step closer on removing us from the requirement of simulation model as input.

wouterjdb:
@anders-kiaer I implemented what you suggested. It works when I run it locally. I'm not entirely satisfied though with the way it has been implemented - it doesn't scale for other keywords. But it might be 'good enough' for now. Maybe you can have a quick look and see whether you see some obvious ways of making the scale-ability better.

The build is currently failing - haven't really looked into the details. There seems to be something fishy with the build script. Also, I noticed that actually the master branch itself has a failed build (that's not great). This is not clear as we don't have a badge for this anymore, that's why I added #17.

src/flownet/parameters/_equilibration.py
```diff
@@ -93,13 +104,42 @@ def render_output(self) -> Dict:
...
+
+        if len(self._unique_eqlnums) > 1:
+            for connections_at_node in self._network.connection_at_nodes:
+                eqlnum_combinations.extend(list(combinations(connections_at_node, 2)))
+
+            eqlnum_combinations = list(set(eqlnum_combinations))
+            eqlnum1 = list(list(zip(*eqlnum_combinations))[0])
+            eqlnum2 = list(list(zip(*eqlnum_combinations))[1])
+            print(eqlnum1)
+            print(eqlnum2)
```
anders-kiaer, line 119: Remove these?
```suggestion
```

examples/norne_parameters.yml
```diff
@@ -4,48 +4,52 @@ flownet:
   data_source:
     eclipse_case: ../tests/data/norne/NORNE_ATW2013
     resample: M
+  pvt:
+    rsvd: ../tests/data/norne/rsvd.csv
```
anders-kiaer, line 8: Put this file instead together with the other ones in `./examples/norne_static`?

examples/norne_parameters.yml
```diff
@@ -4,48 +4,52 @@ flownet:
...
-  fault_tolerance: 1.0e-5
+  fault_tolerance: 0.1
 
 ert:
   static_include_files: ./norne_static
   realizations:
-    num_realizations: 10
-    required_success_percent: 70
+    num_realizations: 1
+    required_success_percent: 1
```
anders-kiaer, line 23: Change this back?

PR #20, opened 2020-04-12
Setting faults as 'None' when the model has no faults

tayloris:
II' used black this time for the code.

When the inputs file has no faults, there is not attribute assigned to the object 'network' with name of 'faults' after network object is declared inside " _run_ahm.py". Therefore the code stops when we ask if "network.faults" has been instantiated
in line 289 isinstance(network.faults, Dict): .

I add a rewrite to this line to ask if the object network has the attribute faults.
if hasattr(network, 'faults'):


src/flownet/network_model/_network_model.py
```diff
@@ -50,6 +50,8 @@ def __init__(
         if isinstance(fault_planes, pd.DataFrame):
             self._fault_planes: pd.DataFrame = fault_planes
             self._faults: Optional[Dict] = self._calculate_faults(fault_tolerance)
+        else:
+            self._faults = None
```
wouterjdb, line 54: Some minor thoughts: maybe, for completeness, use typing in this part of the if-else-statement. Could also a good idea to set _fault_planes to None for completeness. All not strictly necessary, but not doing this in the first place has led to the error you had to fix.

```suggestion
        else:
            self._fault_planes: Optional[pd.DataFrame] = None
            self._faults: Optional[Dict] = None            
```

src/flownet/network_model/_network_model.py
```diff
@@ -50,6 +50,8 @@ def __init__(
         if isinstance(fault_planes, pd.DataFrame):
             self._fault_planes: pd.DataFrame = fault_planes
```
wouterjdb, line 51: As this can be optional, we will then have to change the type annotation:

```suggestion
            self._fault_planes: Optional[pd.DataFrame] = fault_planes
```

src/flownet/network_model/_network_model.py
tayloris: Yes I tried 
` self._faults: Optional[Dict] = None   `
and I work in my computer but I get a problem inside the checking test that the server was doing inside github for running flownet. Something about ` self._faults:` was declared as `Optional[Dict]` before. I'll add this changes to where I get the problem exactly.

tayloris: This is the message given by the server that is runing flownet          
```
Check code style & linting
25s
##[error]Process completed with exit code 1.
Run source $VENV_PATH/bin/activate
All done! ✨ 🍰 ✨
51 files would be left unchanged.

------------------------------------
Your code has been rated at 10.00/10

src/flownet/network_model/_network_model.py:54: error: Attribute '_fault_planes' already defined on line 51
src/flownet/network_model/_network_model.py:55: error: Attribute '_faults' already defined on line 52
src/flownet/network_model/_network_model.py:226: error: Value of type "Optional[Any]" is not indexable
src/flownet/network_model/_network_model.py:235: error: Item "None" of "Optional[Any]" has no attribute "loc"
src/flownet/network_model/_network_model.py:235: error: Value of type "Optional[Any]" is not indexable
Found 5 errors in 1 file (checked 50 source files)
##[error]Process completed with exit code 1.
```

wouterjdb: Maybe this should be the way of writing,

```python
self._fault_planes: Optional[pd.DataFrame] = None
self._faults: Optional[Dict] = None
if isinstance(fault_planes, pd.DataFrame):
            self._fault_planes = fault_planes
            self._faults = self._calculate_faults(fault_tolerance)
```
to solve 
```
src/flownet/network_model/_network_model.py:54: error: Attribute '_fault_planes' already defined on line 51
src/flownet/network_model/_network_model.py:55: error: Attribute '_faults' already defined on line 52
```
Then, to solve,
```
src/flownet/network_model/_network_model.py:226: error: Value of type "Optional[Any]" is not indexable
src/flownet/network_model/_network_model.py:235: error: Item "None" of "Optional[Any]" has no attribute "loc"
src/flownet/network_model/_network_model.py:235: error: Value of type "Optional[Any]" is not indexable
```
you'll have to rewrite

https://github.com/equinor/flownet/blob/0cf8c5ba89ff5136f486f5eb22780e733bf52251/src/flownet/network_model/_network_model.py#L226-L229

to something like:
```python
if self._fault_planes is not None:
    fault_names = self._fault_planes["NAME"].unique().tolist()
        if not fault_names:
            return None
```

**NB.** I didn't test this code; so it might not be completely accurate.

anders-kiaer:
@tayloris Note that the commands that CI is running are these ones:
https://github.com/equinor/flownet/blob/c581f574127acf3ac3597e2b22b44a21f53f1a68/.github/workflows/flownet.yml#L57-L59

By running them locally on your computer you will reduce the time on your feedback loop 🏃 🏎️

tayloris:
I include the changes and now it works :D, I have a question, black and pylint  checks for format but also for checks for ways of programming? (in a more abstract senses)

anders-kiaer:
> I have a question, black and pylint checks for format but also for checks for ways of programming? (in a more abstract senses)

`black` checks only the "style" (i.e. has an opinion on how some given code is written wrt. spaces and line breaks), while `pylint` in our setup checks more "abstract senses" and not "style" (i.e. suggests code simplifications where it sees it etc.).

wouterjdb:
@tayloris If you are ready with your PR and you want a review then you should request a review on the top right of this page (I did the first review on my own initiative but in theory it should be you who asks for it).

tayloris:
How do I proceed  merge the this branch with master? or is it done automatically?

anders-kiaer:
> How do I proceed merge the this branch with master? or is it done automatically?

You push "Squash and merge", however, in order to do so, you need write access. I've tried to give you write access now, but appears I don't have the necessary powers (even though I'm one of the admins on the repo). We'll check internally.

PR #21, opened 2020-04-14
Deploy to PyPI on releases

anders-kiaer:
This finishes the `flownet` part of PyPI deploy.

- [x] `libecl` is already on PyPI.
- [ ] `libres` dependency available on PyPI (in progress by Equinor).
- [ ] `opm-io`  indirect dependency available on PyPI (assigned to Ceetron Solutions).

### DIFF
--- a/.github/workflows/flownet.yml
+++ b/.github/workflows/flownet.yml
@@ -7,6 +7,9 @@ on:
   pull_request:
     branches:
       - master
+  release:
+    types:
+      - published
   
 jobs:
   FlowNet:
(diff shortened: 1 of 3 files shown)


.github/workflows/flownet.yml
```diff
@@ -7,6 +7,9 @@ on:
   pull_request:
     branches:
       - master
+  release:
+    types:
+      - published
```
anders-kiaer, line 12: This will ensure the GitHub worklow also runs on tagged releases in GitHub.

.github/workflows/flownet.yml
```diff
@@ -79,6 +82,16 @@ jobs:
...
+      - name: Build and deploy Python package
+        if: github.event_name == 'release' && matrix.python-version == '3.6'
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.pypi_flownet_token }}
+        run: |
+          python -m pip install --upgrade setuptools wheel twine
+          python setup.py bdist_wheel
+          twine upload dist/*
+
```
anders-kiaer, line 94: When the event triggering the workflow is a release, and we are using Python version 3.6 (just to ensure we are not trying to deploy multiple times), we deploy to PyPI using the standard tool `twine` (together with a `flownet` specific token secret).

PR #44, opened 2020-05-07
Add accuracy metric to ERT workflow

### LINKED ISSUE #29: FlowNet accuracy metric generated using ERT workflows during AHM
Original author: @wouterjdb

We should include the accuracy metric in ERT using workflows during the AHM.

- run after the simulations in an iteration are finished
- output
    - the metrics ({RSME, MSE, ...} for various targets {WOPR, WWPR, WBHP, TOTAL, ...?}) in a single csv file that is appended to
    - a picture that is updated

We should also considering to include the heatmap and/or other useful analytical plots. @edubarrosTNO, which ones were most useful?

Please have a look at to see how you can add a workflow in the code base.

edubarrosTNO:
Resolves #29


src/flownet/ahm/_assisted_history_matching.py
```diff
@@ -270,3 +277,448 @@ def save_iteration_parameters():
...
+        )
+        df_metrics = df_metrics.append(dict_metric_tmp, ignore_index=True)
+
+        # Plotting accuracy metrics over iterations
+        save_plots_metrics(df_metrics, metrics, str_key)
+
+    # Saving accuracy metrics to CSV file
+    df_metrics.to_csv(args.outfile + ".csv", index=False)
+
+    print("[Done]")
```
anders-kiaer, line 724: The additions here does not appear to be tightly dependendt on existing code in `_assisted_history_matching`. Should we move it into a separate module (for readability)?

edubarrosTNO: I can put them in separate file, but they need to be in the same `ahm` module because this workflow is supposed to be run within ERT, thus in the same call of `run_flownet_history_matching` in `ahm._run_ahm.py`.

By the way, just added a commit to fix a typo in the function name (it was `run_flownet_history_mathing` previously).

PR #47, opened 2020-05-13
Remove interpolation of ECLIPSE data and apply time shift

### LINKED ISSUE #36: Time shift in FlowNet results compared to original case
Original author: @wouterjdb 

image

Is this an issue?

olwijn:
Closes #36 . The time shift should make sure that the data enter the schedule as well targets at the correct time. Note that this also moves the observations in time. These should therefore no longer be extracted from the schedule.


anders-kiaer:
Thanks @olwijn. There is a test failing 🔴 that perhaps needs to be updated due to your changes?

PR #48, opened 2020-05-14
Passes realization index as input argument to create_flownet_model

### LINKED ISSUE #41: Same parameter values across ensemble of realizations in prediction mode
When using FlowNet in prediction mode, I observed that the parameters of the realizations of the ensemble (at least the grid parameters - porosity, permeability, etc.) show the same values.

I think something is going wrong when loading the parameters parquet file and writing the (grid) parameter include files. Or maybe something is wrong in history matching mode when saving the parameters into the parquet file.

edubarrosTNO:
Resolves #41


edubarrosTNO:
it took some time to pin-point the root cause of the issue, but it was a simple fix after that: the ERT simulation job `CREATE_FLOWNET_MODEL` is supposed to take `<IENS>` (= realization index) as input, but the template of the ERT configuration file had `<ITER>` (= iteration number) hard-coded as input argument instead

Issue #31, opened 2020-04-24
Implement more tests for pytest coverage to increase from 50 to ...?


anders-kiaer:
Original author: @wouterjdb

PR #52, opened 2020-05-15
Adds more tests for iteration analytics workflow functions

edubarrosTNO:
Contributes to #31

### DIFF
--- a/src/flownet/ahm/_ahm_iteration_analytics.py
+++ b/src/flownet/ahm/_ahm_iteration_analytics.py
@@ -38,39 +38,34 @@ def filter_dataframe(df_in, key_filter, min_value, max_value):
     ]
 
 
-def prepare_opm_reference_data(df, str_key, n):
+def prepare_opm_reference_data(df_opm, str_key, n_real):
     """
     This function extracts data from selected columns of the Pandas dataframe
     containing data from reference simulation, rearranges it into a stacked
-    column vector preserving the original order and repeats it n times to form
-    a matrix for comparison with data from ensemble of n FlowNet simulations
+    column vector preserving the original order and repeats it n_real times to form
+    a matrix for comparison with data from ensemble of n_real FlowNet simulations
 
     Args:
-        df: is the Pandas dataframe containing data from reference simulation
+        df_opm: is the Pandas dataframe containing data from reference simulation
         str_key: is the string to select columns; column names starting with str_key
-        n: is the size of ensemble of FlowNet simulations
+        n_real: is the size of ensemble of FlowNet simulations
 
     Returns:
-        A numpy 2D array [length_data * nb_selected_columns, n] containing data
+        A numpy 2D array [length_data * nb_selected_columns, n_real] containing data
         from selected columns (i.e., quantity of interest for accuracy metric) of
-        reference simulation stacked in a column-vector and replicated into n columns
+        reference simulation stacked in a column-vector and replicated into n_real columns
     """
 
-    keys = df.keys()
-    keys = keys[df.keys().str.contains(str_key)]
-    data = df[keys].values
+    keys = df_opm.keys()
+    keys = keys[df_opm.keys().str.contains(str_key)]
+    data = df_opm[keys].values
     data = data.flatten()
-    #data = np.reshape(
-    #    df[keys].values,
-    #    (df[keys].values.shape[0] * df[keys].values.shape[1], 1),
-    #    order="F",
-    #)
-    data = np.transpose(np.tile(data, (n, 1)))
+    data = np.transpose(np.tile(data, (n_real, 1)))
 
     return data
 
 
-def prepare_flownet_data(df, str_key, n):
+def prepare_flownet_data(df_flownet, str_key, n_real):
     """
     This function extracts data from selected columns of the Pandas dataframe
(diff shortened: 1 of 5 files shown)


edubarrosTNO:
In the process of adding more tests for the functions introduced for analytics workflow, I observed that `normalize_data` function had a small issue. This is now fixed

edubarrosTNO:
Added some more tests: we are now at 55.48%

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -38,60 +38,57 @@ def filter_dataframe(df_in, key_filter, min_value, max_value):
...
-        df[keys].values,
-        (df[keys].values.shape[0] * df[keys].values.shape[1], 1),
-        order="F",
-    )
-    data = np.tile(data, n)
+    keys = df_opm.keys()
+    keys = keys[df_opm.keys().str.contains(str_key)]
+    data = df_opm[keys].values
+    data = data.flatten()
+    data = np.transpose(np.tile(data, (n_real, 1)))
```
anders-kiaer, line 63: Have not tested, but could this be simplified to e.g.
```suggestion
    data = np.transpose(np.tile(df_opm[keys].values.flatten(), (n_real, 1)))
```

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -38,60 +38,57 @@ def filter_dataframe(df_in, key_filter, min_value, max_value):
...
         to one realization of the ensemble
     """
 
-    keys = df.keys()
-    keys = keys[df.keys().str.contains(str_key)]
-    data = df[keys].values
+    keys = df_flownet.keys()
+    keys = keys[df_flownet.keys().str.contains(str_key)]
+    data = df_flownet[keys].values
     data = data.flatten()
```
anders-kiaer, line 90:
```suggestion
    data = df_flownet[keys].values.flatten()
```

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -38,60 +38,57 @@ def filter_dataframe(df_in, key_filter, min_value, max_value):
...
-    data = np.reshape(data, (int(data.shape[0] / n), n), order="F")
+    data = np.reshape(data, (int(data.shape[0] / n_real), n_real), order="F")
```
anders-kiaer, line 91:
```suggestion
    data = np.reshape(data, (data.shape[0] // n_real, n_real), order="F")
```

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -296,15 +294,18 @@ def make_dataframe_simulation_data(path, eclbase_file, keys):
...
-        df_tmp["DATE"] = pd.Series(dates)
-        df_tmp["REAL_ID"] = pd.Series(id_real * np.ones(len(dates)), dtype=int)
+            df_sim["DATE"] = pd.Series(realizations_dict[runpath].dates)
+            df_sim["REAL_ID"] = pd.Series(
+                id_real * np.ones(len(realizations_dict[runpath].dates)), dtype=int
+            )
+        df_tmp["DATE"] = pd.Series(realizations_dict[runpath].dates)
+        df_tmp["REAL_ID"] = pd.Series(
+            id_real * np.ones(len(realizations_dict[runpath].dates)), dtype=int
+        )
```
anders-kiaer, line 308: What is the reasoning behind not creating the variable `dates` (which is used some times in this code block)?

tests/test_calculation_accuracy_metric.py
```diff
@@ -1,11 +1,13 @@
...
 def test_calculation_accuracy_metric() -> None:
     assert (
         accuracy_metric(
-            [[0.5, 0.5, 0.5], [0.5, 0.5, 0.5]],
-            [[0.4, 0.4, 0.4], [0.6, 0.6, 0.6]],
+            np.array([[0.5, 0.5, 0.5], [0.5, 0.5, 0.5]]),
+            np.array([[0.4, 0.4, 0.4], [0.6, 0.6, 0.6]]),
             "RMSE",
         )
         == 0.09999999999999998
```
anders-kiaer, line 13: Maybe this test should use something like [`math.isclose`](https://docs.python.org/dev/library/math.html#math.isclose) in order to not get a false negative due to floating point issues?

src/flownet/ahm/_ahm_iteration_analytics.py
edubarrosTNO: `pylint` was complaining of "too many locals"

tests/test_calculation_accuracy_metric.py
edubarrosTNO: good point! I have never used this, but will take a look into it

src/flownet/ahm/_ahm_iteration_analytics.py
anders-kiaer: Ah, ok. Maybe we can keep it, and simply (at least for now) disable that pylint message in this function? 🙂 
http://pylint.pycqa.org/en/latest/faq.html#is-it-possible-to-locally-disable-a-particular-message

PR #58, opened 2020-05-24
Shift observations used by ERT

anders-kiaer:


wouterjdb:
@anders-kiaer Is this PR still being worked on?

anders-kiaer:
Not atm, but I think it is still relevant.

PR #74, opened 2020-06-22
Optimization of parameters using hyperopt

wouterjdb:
This PR explores optimization of uncertain parameters using an optimization packages, in this case hyperopt. The aim is to generate a single optimized model with a low evaluation value (RMSE?) for both historic and prediction periods.

This would be a single optimized realization and looses the concept of uncertainty. Another open question is how to add uncertainty back in the picture.

This PR is just there to explore if this could work - I'm not yet aiming for a merge.


wouterjdb:
Currently running a first test with **sum(abs(FOPT-FOPTH))** as objective function on a Norne model with no additional nodes.

The following image shows the iteration number vs the loss. Clearly until about 600 iterations hyperopt is pushing the average down - seems like it has become stagnant now.

image

Of course there are much better ways to design the loss function.

To be continued...

wouterjdb:
All 1000 runs are now done; no further improvement. Results of the best run (#5xx):

image

PR #76, opened 2020-06-24
Grab and print ERT error message if ERT fails

### LINKED ISSUE #49: Investigate failed runs in cloud. Print to console?

anders-kiaer:
Closes #49.

### DIFF
--- a/src/flownet/ahm/_assisted_history_matching.py
+++ b/src/flownet/ahm/_assisted_history_matching.py
@@ -120,15 +120,25 @@ class AssistedHistoryMatching:
                 )
             )
 
-        # Ignore deprecation warnings (ERT as of August 2019 has a lot of them
-        # due to transition to Python3)
-        subprocess.run(
-            "export PYTHONWARNINGS=ignore::DeprecationWarning;"
-            f"ert es_mda --weights {','.join(map(str, weights))!r} ahm_config.ert",
-            cwd=self.output_folder,
-            shell=True,
-            check=True,
-        )
+        try:
+            # Ignore deprecation warnings (ERT as of August 2019 has a lot of them
+            # due to transition to Python3)
+            subprocess.run(
+                "export PYTHONWARNINGS=ignore::DeprecationWarning;"
+                f"ert es_mda --weights {','.join(map(str, weights))!r} ahm_config.ert",
+                cwd=self.output_folder,
+                shell=True,
+                check=True,
+            )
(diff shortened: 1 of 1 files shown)


src/flownet/ahm/_assisted_history_matching.py
```diff
@@ -120,15 +120,25 @@ def run_ert(self, weights: List[float]):
...
+                cwd=self.output_folder,
+                shell=True,
+                check=True,
+            )
+        except subprocess.CalledProcessError:
+            error_files = glob.glob(
+                str(
+                    self.output_folder
+                    / self._ert_config["runpath"].replace("%d", "*")
+                    / "ERROR"
```
wouterjdb, line 138: Will this give enough info - don't we need the *stderr files?

anders-kiaer: With the current code in this PR  you already get `stderr` information (and more):
```
<error>
  <time>07:28:04</time>
  <job>FLOW</job>
  <reason>Process exited with status code 1</reason>
  <stderr>
No suitable configuration found, valid are Twophase, polymer, foam, brine, solvent, energy, blackoil.
Traceback (most recent call last):
  File "/home/runner/work/flownet/flownet/flownet_venv/share/ert/forward-models/res/script/flow", line 6, in <module>
    run(config, [arg for arg in sys.argv[1:] if len(arg) > 0])
  File "/home/runner/work/flownet/flownet/flownet_venv/lib/python3.6/dist-packages/res/fm/ecl/script.py", line 22, in run
    run.runEclipse()
  File "/home/runner/work/flownet/flownet/flownet_venv/lib/python3.6/dist-packages/res/fm/ecl/ecl_run.py", line 215, in runEclipse
    raise Exception("The eclipse executable:%s exited with error status: %d" % (self.sim.executable, return_code))
Exception: The eclipse executable:/usr/bin/flow exited with error status: 1
</stderr>
  <stderr_file>/home/runner/work/flownet/flownet/examples/some_ahm_run/output/runpath/realization-1/iter-0/FLOW.stderr.5</stderr_file>
</error>
```

I think what however could be a slight "problem" is to distinguish between a realization/job that randomly fails (here we pick the first `ERROR`), and the more useful output of something that fails consistently across realizations (which could happen later) if they reach that job at all. I suggest we wait doing something on that until it appears to be a problem in practice - when done it should anyway fit nicely with the suggested structure here.

PR #77, opened 2020-06-25
WSTAT warning and default well status

### LINKED ISSUE #69: wells status  by default
Flownet set wells status "CLOSED" by default.  I'm applying flownet for the egg model and all wells are set to be "CLOSED" by default. Maybe, flownets should warn the user when wells status is not defined in the original model.

tayloris:
Closes #69. 

I include a warning message to let the user know well status are set OPEN by default.

However the message appears at the beginning of the run and I don't like  the way is presented, so feel free to comment on that. we can discuss it on the meeting tomorrow.


tayloris:
I'll do the code styling later

anders-kiaer:
This looks good @tayloris 🎉 

As discussed, a small minor change suggested, then this is ready for merge 🚀 
> If no `WSTAT` values exist, then we have default `OPEN`, if _some_ `WSTAT` values exists (`!= np.nan) then we keep the default (`STOP`) as of now.

PR #78, opened 2020-06-26
I62 wstat

### LINKED ISSUE #62: Production well BR-P-20 not shutting in
While other producing wells are correctly shutting in at high WCT, this particular well (the final production well) does not shutin in in FLOWNET.

olwijn:
Closes #62.

The default type for production wells is set to SHUT instead of None to ensure that schedule entries rae created also after shutting in the well.


olwijn:
Some (not all) Norne runs are failing due to convergence problems. The problems arise at different times and for different wells in different FlowNet realizations.

Issue #61, opened 2020-06-03
Error in flownet_save_iteration_analytics in CI/CD build


wouterjdb:
In a successful github workflow run one finds the following error in the logs:

```
The script 'ExternalErtScript' caused an error while running:

Simulations completed.
Traceback (most recent call last):
  File "/home/runner/work/flownet/flownet/flownet_venv/lib/python3.7/site-packages/pandas/core/indexes/base.py", line 2897, in get_loc
    return self._engine.get_loc(key)
  File "pandas/_libs/index.pyx", line 107, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/index.pyx", line 131, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 1607, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 1614, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'DATE'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/runner/work/flownet/flownet/flownet_venv/bin/flownet_save_iteration_analytics", line 8, in <module>
    sys.exit(save_iteration_analytics())
  File "/home/runner/work/flownet/flownet/flownet_venv/lib/python3.7/site-packages/flownet/ahm/_ahm_iteration_analytics.py", line 420, in save_iteration_analytics
    df_sim = df_sim[df_sim["DATE"].isin(df_obs["DATE"])]
  File "/home/runner/work/flownet/flownet/flownet_venv/lib/python3.7/site-packages/pandas/core/frame.py", line 2995, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/home/runner/work/flownet/flownet/flownet_venv/lib/python3.7/site-packages/pandas/core/indexes/base.py", line 2899, in get_loc
    return self._engine.get_loc(self._maybe_cast_indexer(key))
  File "pandas/_libs/index.pyx", line 107, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/index.pyx", line 131, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 1607, in pandas._libs.hashtable.PyObjectHashTable.get_item
[...]

PR #81, opened 2020-07-01
Fixes bug in analysis workflow

edubarrosTNO:
Partially addresses #61.

A new way of loading data from the reference simulation (OPM-Flow/Eclipse) is added, using functionality from `flownet.data.from_eclipse` instead of the previous approach of fetching data from the observation file into dictionary which was causing the random order behavior

### DIFF
--- a/src/flownet/ahm/_ahm_iteration_analytics.py
+++ b/src/flownet/ahm/_ahm_iteration_analytics.py
@@ -11,9 +11,11 @@ from matplotlib.ticker import ScalarFormatter
 import numpy as np
 import pandas as pd
 from sklearn.metrics import mean_squared_error, mean_absolute_error, r2_score
-import yaml
 
 from ecl.summary import EclSum
+
+from ..config_parser import parse_config
+from ..data import EclipseData
 
 
 def filter_dataframe(df_in, key_filter, min_value, max_value):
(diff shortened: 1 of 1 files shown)


edubarrosTNO:
`flownet.data.from_eclipse` currently only loads well data (e.g. WBHP, WOPR, …). If we would like to add quantities for accuracy analysis that also include field data (e.g. FOPR, FWPR, …) that method will have to be updated. Thus, for now, the analysis workflow will only support well data quantities (provided that they are requested in the SUMMARY section of the FlowNet simulations and of the OPM-Flow/Eclipse reference simulation)

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -287,6 +258,7 @@ def make_dataframe_simulation_data(path, eclbase_file, keys):
     # Load summary files of latest iteration
     # with concurrent.futures.ProcessPoolExecutor() as executor:
     for runpath, eclbase in list(zip(runpath_list, itertools.repeat(eclbase_file))):
+        print(runpath, eclbase)
```
anders-kiaer, line 261: Remove?

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -397,39 +370,55 @@ def save_iteration_analytics():
...
+    df_obs["DATE"] = pd.to_datetime(df_obs["DATE"])
+    df_sim["DATE"] = pd.to_datetime(df_sim["DATE"], format="%Y-%m-%dT%H:%M:%S")
 
     # Filter dataframe base on measurement dates
+    df_obs = df_obs[df_obs["DATE"].isin(df_sim["DATE"])]
     df_sim = df_sim[df_sim["DATE"].isin(df_obs["DATE"])]
 
+    print(df_obs["DATE"].values)
+    print(df_sim["DATE"].values)
+    print("compare!\n")
```
anders-kiaer, line 398: Remove?

PR #91, opened 2020-07-09
Update GitHub actions workflow

anders-kiaer:
GitHub cache is working in their latest action version.

Astroid/Pylint is also fixed upstream - bumping to latest version (which also uncovers some small issues with f-strings).


wouterjdb:
Passed without cache. But crashed on pytest during the round where a cache was used on not being able to find `norne_parameters.yml`.

anders-kiaer:
Looks like it fails due to missing input case data... which again is due to this line being part of `build_environment.sh` (which is not re-run when cache exists):
https://github.com/equinor/flownet/blob/ec652b16c2da811978cd57a158297110b275b20b/build_environment.sh#L32

That line should probably be taken out of the `build_environment.sh` script, as it is not part of "building the Python environment". The test cases/models should soon also come from another repository (https://github.com/equinor/flownet-testdata) and not this one.

PR #92, opened 2020-07-09
Correct shutting in of last production well

### LINKED ISSUE #62: Production well BR-P-20 not shutting in

olwijn:
Closes #62 and Closes #78


olwijn:
In the Norne deck some wells are never producing or injecting. The type of these wells can therefore not be determined from the ECLIPSE output. In the FlowNet workflow their type and phase are by default set to None. However, a backwards infill procedure is applied that resets these None values to the values for the subsequent well in the list. This seems rather arbitrary.

olwijn:
The assignment of well PHASE and TYPE and filling of None entries is moved to the dataframe construction for each individual well. This better ensures that the values appropriate for that specific well are used instead of the values for the next well in the overall well list. 

Some wells in the Norne deck never become active. At this moment they are assigned the type and phase of the next well (or previous one if it is the very last well), which may not be correct. I have kept this in such these wells remain in the schedule in case they provide pressure measurements.

PR #95, opened 2020-07-13
Add early check for training set definition

### LINKED ISSUE #27: Training fraction and end date allowed at the same time in config
Original authour: @wouterjdb 

If you set both the training set fraction and end date you'll get the following warning when running FlowNet:

```
Training set fraction and training set end date are both defined in config file.
The input given for training set fraction will be ignored.
The training set end date will be used to calculate the training set fraction.
```

I'd actually suggest to let the config parser handle this and stop execution early on. A warning during execution is a bit 'too silent' if you ask me.

wouterjdb:
Close #27


src/flownet/ahm/_run_ahm.py
```diff
@@ -37,10 +37,8 @@ def _find_training_set_fraction(
 
     if config.flownet.training_set_end_date is not None:
         if config.flownet.training_set_fraction is not None:
-            print(
+            raise ValueError(
```
anders-kiaer, line 40: Is this check necessary when failing earlier?

src/flownet/config_parser/_config_parser.py
```diff
@@ -518,4 +518,10 @@ def _to_abs_path(path: Optional[str]) -> str:
         ):
             raise ValueError(f"The input case {suffix} file does not exist")
 
+    if config.flownet.training_set_end_date and config.flownet.training_set_fraction:
+        raise ValueError(
+            "Ambiguous configuration input: 'training_set_fraction' and 'training_set_end_date' are "
+            "mutually exclusive but are currently both defined in the configuration file."
```
anders-kiaer, line 524: Can we simplify message and remove `mutually exclusive but`?

src/flownet/ahm/_run_ahm.py
wouterjdb: In theory it should never get there.... but if one bypasses the config file and calls functions directly it will trigger. 

So, yeah, we could remove it if we only go via the config file route.

PR #106, opened 2020-08-06
Fixed bug where the config would be parsed incorrectly when no aquifer is defined

### LINKED ISSUE #105: Config file parsed incorrectly when no aquifer is defined
Reported by @tayloris.

The following error is thrown when a config is parsed without an aquifer:

```
Traceback (most recent call last):
  File "/home/manuel/repos/projects/Flownet_August2020/flownet/venv/bin/flownet", line 11, in <module>
    load_entry_point('flownet', 'console_scripts', 'flownet')()
  File "/home/manuel/repos/projects/Flownet_August2020/flownet/src/flownet/_command_line.py", line 206, in main
    args.func(args)
  File "/home/manuel/repos/projects/Flownet_August2020/flownet/src/flownet/_command_line.py", line 66, in flownet_ahm
    config = parse_config(args.config)
  File "/home/manuel/repos/projects/Flownet_August2020/flownet/src/flownet/config_parser/_config_parser.py", line 557, in parse_config
    "Ambiguous configuration input: 'size_in_bulkvolumes' in 'aquifer' needs to be defined using "
ValueError: Ambiguous configuration input: 'size_in_bulkvolumes' in 'aquifer' needs to be defined using 'min', 'max' and 'log_unif'. Currently one or more parameters are missing.
```

The following if-statement should be changed,

https://github.com/equinor/flownet/blob/57b10f02f967b050289b7cfe447b2d5d88cbba48/src/flownet/config_parser/_config_parser.py#L555

to
```python
if all(config.model_parameters.aquifer[0:3]) and not all(config.model_parameters.aquifer.size_in_bulkvolumes):
```

This then leads to a further error:
```
Adding aquifer nodes...Traceback (most recent call last):
  File "/home/manuel/repos/projects/Flownet_August2020/flownet/venv/bin/flownet", line 11, in <module>
    load_entry_point('flownet', 'console_scripts', 'flownet')()
  File "/home/manuel/repos/projects/Flownet_August2020/flownet/src/flownet/_command_line.py", line 206, in main
    args.func(args)
  File "/home/manuel/repos/projects/Flownet_August2020/flownet/src/flownet/_command_line.py", line 67, in flownet_ahm
    run_flownet_history_matching(config, args)
  File "/home/manuel/repos/projects/Flownet_August2020/flownet/src/flownet/ahm/_run_ahm.py", line 157, in run_flownet_history_matching
    df_connections: pd.DataFrame = create_connections(df_coordinates, config)
  File "/home/manuel/repos/projects/Flownet_August2020/flownet/src/flownet/network_model/_generate_connections.py", line 462, in create_connections
    starts, ends, scheme, fraction=fraction, delta_depth=delta_depth
  File "/home/manuel/repos/projects/Flownet_August2020/flownet/src/flownet/network_model/_generate_connections.py", line 399, in _generate_aquifer_connections
    0 : math.floor(fraction * len(all_coordinates_sorted))
TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'

```

as a result of 

https://github.com/equinor/flownet/blob/57b10f02f967b050289b7cfe447b2d5d88cbba48/src/flownet/network_model/_generate_connections.py#L456

not working as intended.

This line will need to be changed to:

```python
if all(aquifer_config[0:3]) and all(aquifer_config.size_in_bulkvolumes):
```

Then it should be OK again.

wouterjdb:
Closes #105

### DIFF
--- a/src/flownet/network_model/_generate_connections.py
+++ b/src/flownet/network_model/_generate_connections.py
@@ -453,7 +453,7 @@ def create_connections(
     aquifer_ends: List[Coordinate] = []
 
     aquifer_config = configuration.model_parameters.aquifer
-    if aquifer_config is not None:
+    if all(aquifer_config[0:3]) and all(aquifer_config.size_in_bulkvolumes):
         scheme = aquifer_config.scheme
         fraction = aquifer_config.fraction
         delta_depth = aquifer_config.delta_depth
--- a/src/flownet/config_parser/_config_parser.py
+++ b/src/flownet/config_parser/_config_parser.py
@@ -552,7 +552,9 @@ def parse_config(configuration_file: pathlib.Path) -> ConfigSuite.snapshot:
             "'scheme', 'fraction', 'delta_depth' and a 'size_in_bulkvolumes' distribution ('min', 'max', 'logunif')."
             "Currently one or more parameters are missing."
         )
-    if not all(config.model_parameters.aquifer.size_in_bulkvolumes):
+    if all(config.model_parameters.aquifer[0:3]) and not all(
+        config.model_parameters.aquifer.size_in_bulkvolumes
+    ):
         raise ValueError(
             "Ambiguous configuration input: 'size_in_bulkvolumes' in 'aquifer' needs to be defined using "
             "'min', 'max' and 'log_unif'. Currently one or more parameters are missing."


src/flownet/network_model/_generate_connections.py
```diff
@@ -453,7 +453,7 @@ def create_connections(
     aquifer_ends: List[Coordinate] = []
 
     aquifer_config = configuration.model_parameters.aquifer
-    if aquifer_config is not None:
+    if all(aquifer_config[0:3]) and all(aquifer_config.size_in_bulkvolumes):
```
olelod, line 456: would it be an idea to add a configuration.flownet.aquifer boolean (like with fast_pyscal) so you can just have `if configuration.flownet.aquifer:` here?

wouterjdb: No, because that would mean you have to define the aquifer at two spots.

I fully agree with you that this if-statement is ugly. And I have now the same issue in other places as well.... I asked @markusdregi if there is a better solution for this.

olelod: Do you really? Can't you just set it false as default, and set it to true if the parsing of the aquifer parameter passes all checks?

markusdregi: There is no better solution for this as of now, but I've been considering for some time the possibility of supporting derived data, that cannot be configured by a user, but that will be computed based on the input configuration. That could fix your "problem" here.

Also, you can of course implement derived data in the current system (basically what Ole suggests above); but that will have the rather odd consequence of exposing a keyword to the user that the user will have no say over. She can flip it one way or the other, and it will make no difference as the schema will always set it to the correct value.

PR #111, opened 2020-08-07
Phases in simulation model

### LINKED ISSUE #39: Phases in simulation model
Original author: @olelod 

Currently the phases of the model are extracted from the relative permeability given. 

        if self._swof and self._sgof:
            str_runspec_section = "OIL\nWATER\nGAS\n"
        elif self._swof:
            str_runspec_section = "OIL\nWATER\n"
        elif self._sgof:
            str_runspec_section = "OIL\nGAS\n"

        return {
            "RUNSPEC": str_runspec_section,
            "REGIONS": write_grdecl_file(merged_df_satnum, "SATNUM", int_type=True),
            "PROPS": str_props_section,
        }

Should we reverse this and make the user provide which phases are present in the config yaml? This would make it easier to check if the input needed is provided (relperm, contacts etc.).

olelod:
Closes #39 

User provide a list of phases in the flownet part of the config .yml file. 

The parsing of the config file checks for valid phases, the required model parameters for relative permeability etc. More checks could surely be added.


src/flownet/config_parser/_config_parser.py
```diff
@@ -459,6 +463,112 @@ def _to_abs_path(path: Optional[str]) -> str:
 
     config = suite.snapshot
 
+    req_relp_parameters: List[str] = []
+
+    for phase in config.flownet.phases:
+        if phase.lower() not in ["oil", "gas", "water", "disgas", "vapoil"]:
```
anders-kiaer, line 469: There are a lof of `.lower()` around. Maybe we could just transform it to lower on initial parse with `ConfigSuite`? (Similarly to how e.g. relative paths are converted to absolute paths). That would the `if` statements.

src/flownet/config_parser/_config_parser.py
```diff
@@ -480,13 +480,13 @@ def _to_abs_path(path: Optional[str]) -> str:
                 f"The valid phases are 'oil', 'gas', 'water', 'disgas' and 'vapoil'"
             )
     if any(
-        x in [y for y in config.flownet.phases] for x in ["vapoil", "disgas"]
-    ) and not all(x in [y for y in config.flownet.phases] for x in ["oil", "gas"]):
+        x in config.flownet.phases for x in ["vapoil", "disgas"]
+    ) and not all(x in config.flownet.phases for x in ["oil", "gas"]):
```
anders-kiaer: Maybe `set` notation is easier to read here (as the current logic might take some time to understand)?

Not tested, but in that case something like
```python
if not set(["vapoil", "disgas"]).isdisjoint(config.flownet_phases) and not set(["oil", "gas"]).issubset(config.flownet.phases):
```
If so, the other `if` sentences having ~the same structure should also move to `set`.

PR #112, opened 2020-08-11
Add FlowNet Azure Function Docker container

wouterjdb:
This PR adds the possibility of executing the Flow simulation and Render Realization forward models in the azure cloud.


wouterjdb:
The current implementation of the Azure Functions works fine. However, the aim is to run all realizations in parallel. With the current implementation this would mean that one has to set the LOCAL queue max running option to the number of realizations. However, as ERT will spawn threads for each realization the local machine will potentially freeze (observed on Ubuntu 18.04 LTS) due to to many threads waiting on one-another.

That limits the usefulness of calling the Azure Functions in this way significantly.

A potential solution is to run ERT with a single thread and let ERT only run a PRE_SIMULATION hooked workflow which calls an _Azure Durable Function_ that takes care of the parallelization.
Another option is to use another ERT queue system that is hosted in Azure. This has the benefit of not breaking the standard ERT config file between local and cloud runs.

Further development is currently on hold to make a decision on which of these routes to choose.

wouterjdb:
New PR made.

PR #123, opened 2020-08-19
Speed up ray tracing

### LINKED ISSUE #102: Speed-up and/or reuse of fault ray tracing results

wouterjdb:
Close #102 

The ray tracing is now performed:

1) Loop through all faults
2) Perform a triangulation of all points belonging to a fault plane and store the triangles
3) For each connection, find all triangles in its bounding box, perform ray tracing using the Möller-Trumbore intersection algorithm.
4) If an intersection is found, identify the grid blocks that are associated with the intersection.

The process has now been reduced in time from multiple minutes to seconds (depending on the config settings).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,6 @@ REQUIREMENTS = [
     "pyyaml>=5.2",
     "scikit-learn>=0.22",
     "scipy>=1.4",
-    "tqdm>=4.43",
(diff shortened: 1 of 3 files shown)


anders-kiaer (approved):
🎉🏎️🏇

PR #127, opened 2020-08-21
Enables the prescription of a distribution mean from a previous HM

### LINKED ISSUE #66: Use posterior ensemble mean for new prior
Practical experience shows that history matching results may still contain biases. A two-step solution approach was proposed by Guo and Reynolds (2019) to address this, in which the posterior ensemble means are used as the means for a new prior in a second application of the history matching workflow.

If the ensemble of parameter estimates is saved in a compact form at the end of the history matching, new functionality could be introduced to specify that the mean of the new prior should be taken from this file.

olwijn:
Closes #66 

The functionality is activated by adding the keyword ahm_case to the model_parameters section of the config:

image

The specified folder is currently expected to contain the file parameters_iteration-latest.parquet.gzip


src/flownet/ahm/_run_ahm.py
```diff
@@ -119,6 +122,94 @@ def _get_distribution(
...
+        n = len(parameter.random_variables)
+
+        for i, var in enumerate(parameter.random_variables):
+            mean = parameter.mean_values[i]
+
+            loguniform = var.ert_gen_kw.split()[0] == "LOGUNIF"
+            dist_min0 = var.minimum
+            dist_max0 = var.maximum
+
+            if loguniform is True:
```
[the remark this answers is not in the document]
olwijn: I have now created a separate function to determine min/max

src/flownet/ahm/_run_ahm.py
```diff
@@ -119,6 +122,94 @@ def _get_distribution(
     return df
 
 
+def update_distribution(parameters, ahm_case):
```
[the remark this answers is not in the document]
olwijn: type annotation was added

olwijn:
image

I don't quite understand the problem. reveal_type(parameters) on input and output gives:

image

wouterjdb:
Mypy is complaining about that you tried to get attributes from something which is allowed to be a str, according to the input type,

`parameters: Union[str, List[str]]`

and a string does not have these attributes.

olwijn:
Two more changes are needed to fully integrate this functionality: (1) add a test to workflow.yml in .github/workflows : add a second history match run with restart after the first one and before the prediction run, (2) do not rebuild the flownet when using the restart, but instead use the functionality used for prediction runs to read in the flownet.

olwijn:
Two more changes are needed to fully integrate this functionality: (1) add a test to workflow.yml in .github/workflows : add a second history match run with restart after the first one and before the prediction run, (2) do not rebuild the flownet when using the restart, but instead use the functionality used for prediction runs to read in the flownet.

.github/workflows/flownet.yml
```diff
@@ -88,6 +88,7 @@ jobs:
           flownet ahm ./config/assisted_history_matching.yml ./some_ahm_run --update-config ./ci_config/assisted_history_matching.yml
```
anders-kiaer, line 88: Can we add one line here where the new functionality is used? Then we are sure the added functionality runs :green_heart: :slightly_smiling_face:

src/flownet/_command_line.py
```diff
@@ -144,6 +144,13 @@ def main():
         help="Optional configuration file which values will update the main config. "
         "Any relative paths in this file will also be assumed relative to the main config.",
     )
+    parser_ahm.add_argument(
+        "--restart-folder",
+        type=pathlib.Path,
+        default=None,
+        help="Optional folder containing results from a previous history match. "
+        "Any relative paths in this file will also be assumed relative to the main config.",
```
wouterjdb, line 152:
```suggestion
        help="Optional folder containing results from a previous history match.",
```

src/flownet/ahm/_assisted_history_matching.py
```diff
@@ -94,7 +94,7 @@ def run_ert(self, weights: List[float]):
         Args:
             weights: Weights for the iterated ensemble smoother to use.
 
-        Returns:
+        Returns:self.output_folder
```
wouterjdb, line 97:
```suggestion
        Returns:
```

wouterjdb (changes requested):
LGTM - two minor things. You can change those from within your browser (accept the suggestions). Approved afterward. 👍

src/flownet/ahm/_run_ahm.py
```diff
@@ -1,6 +1,10 @@
 import argparse
 from typing import Union, List, Optional
```
anders-kiaer, line 3:
```suggestion

```

anders-kiaer (commented):
One minor suggestion, otherwise LGTM as well.

wouterjdb (approved):

PR #128, opened 2020-08-21
Add option to use model hull to steer node/connection generation

### LINKED ISSUE #124: Hull from simulation model
Currently we use a hull that is created based well locations (left image is an example):

image

What strikes is that in the Norne model we have two area in the model, including wells, that have non-reservoir in between. It would make it somewhat easier to train a model if we do not include nodes in the non-reservoir zone which would require to alter the hull.

I suggest that we add an option that would allow for taking the shell of the model as hull, this would make it easier to recreate a Norne model geometry in FlowNet. I also think that this is still inline with the data-driven character of FlowNet as from seismic and/or a basic geo model one can get this information as well.

wouterjdb:
Closes #124.

Three new options can be specified in the config:

```yaml
data_sources:
  concave_hull: True
```

Which limits new nodes to the volume of the original simulation model.
And,

```yaml
flownet:
  max_distance: 1000
  max_distance_fraction: 0.15
```

Which limits the max length of connections after doing the Delaunay triangulation which created the connections.

The resulting FlowNet model for Norne now looks like:
image


tests/test_generate_connections.py
```diff
@@ -0,0 +1,115 @@
...
+
+    starts, ends = _generate_connections(
+        df_coordinates=DF_COORDINATES,
+        configuration=config,
+        concave_hull_bounding_boxes=np.array([0, 2, 0, 2, 0, 2]).reshape(-1, 6),
+    )
+
+    assert len(starts) == len(ends)
+    assert starts is not STARTS
+    assert ends is not ENDS
```
anders-kiaer, line 84: Worth making sure we don't get into float problems? E.g.  https://docs.pytest.org/en/latest/reference.html#pytest-approx?

anders-kiaer (approved):
One suggestion to consider, otherwise LVGTM! 🚀🌻

tests/test_generate_connections.py
wouterjdb: Good point, will do.

PR #136, opened 2020-08-24
AllowNone for queue name

### LINKED ISSUE #135: Queue "name: ..." needs to be specified in the config
The

```yaml
ert:
  queue:
     name: ...
```

needs to be specified in the configuration file otherwise FlowNet will fail while parsing the yaml file. This is not the behavior we would like to see as the queue name is even ignored when the LOCAL queue is used --> It should be allowed to not specify it at all.

wouterjdb:
Closes #135


tayloris (approved):
Good

PR #144, opened 2020-08-26
Automatically create documentation from schema

anders-kiaer:
Waiting on feedback on https://github.com/equinor/configsuite/pull/157.

### DIFF
--- a/src/flownet/config_parser/_config_parser.py
+++ b/src/flownet/config_parser/_config_parser.py
@@ -1,13 +1,19 @@
 import os
 import pathlib
-from typing import Dict, Optional, List
+from typing import Dict, Optional, List, Union
 
 import yaml
 import configsuite
 from configsuite import types, MetaKeys as MK, ConfigSuite
 
 
-def create_schema(_to_abs_path) -> Dict:
+# Small workaround while waiting for https://github.com/equinor/configsuite/pull/157
+# to be merged and released in upstream ConfigSuite:
+def create_schema_without_arguments() -> Dict:
+    return create_schema()
+
+
+def create_schema(config_folder: Optional[pathlib.Path] = None) -> Dict:
     """
     Returns a configsuite type schema, where configuration value types are defined, together
     with which values are optional and/or has default values.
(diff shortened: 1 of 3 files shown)


wouterjdb:
Blocked because you are waiting on a PR in configsuite? Could you supply the link such that we can follow the status here?

anders-kiaer:
You are too quick ;)

wouterjdb:
See example for `MK.description`: https://github.com/equinor/configsuite/blob/412dfd7877fa5e23fea7529d66cf17969ad242b0/tests/data/pets.py

src/flownet/config_parser/_config_parser.py
```diff
@@ -19,10 +25,40 @@ def create_schema(_to_abs_path) -> Dict:
...
     """
+
+    @configsuite.transformation_msg("Convert string to lower case")
+    def _to_lower(input_data: Union[List[str], str]) -> Union[List[str], str]:
+        if isinstance(input_data, str):
+            return input_data.lower()
+
+        return [x.lower() for x in input_data]
+
+    @configsuite.transformation_msg("Tries to convert input to absolute path")
```
wouterjdb, line 36: Always wondered about why this is so vaguely written. Why not "Convert path to absolute path" - or something like this.

It is not really trying - it's doing it otherwise it crashes....`?

src/flownet/config_parser/_config_parser.py
```diff
@@ -35,9 +71,11 @@ def create_schema(_to_abs_path) -> Dict:
                                     "input_case": {
                                         MK.Type: types.String,
                                         MK.Transformation: _to_abs_path,
+                                        MK.Description: "Eclipse input case to be used as data source for FlowNet",
```
wouterjdb, line 74:
```suggestion
                                        MK.Description: "Simulation input case to be used as data source for FlowNet",
```

src/flownet/config_parser/_config_parser.py
```diff
@@ -182,14 +222,26 @@ def create_schema(_to_abs_path) -> Dict:
...
                     "perforation_handling_strategy": {
                         MK.Type: types.String,
                         MK.Default: "bottom_point",
                     },
-                    "fast_pyscal": {MK.Type: types.Bool, MK.Default: True},
+                    "fast_pyscal": {
+                        MK.Type: types.Bool,
+                        MK.Default: True,
+                        MK.Description: "If True, pyscal uses some reasonable "
+                        "approximation to reduce running time.",
```
wouterjdb, line 243: Also skips validity tests.

src/flownet/config_parser/_config_parser.py
```diff
@@ -303,6 +377,9 @@ def create_schema(_to_abs_path) -> Dict:
                     },
                     "bulkvolume_mult": {
                         MK.Type: types.NamedDict,
+                        MK.Description: "Description of bulk volume multiplier "
```
wouterjdb, line 380: Should this describe a bit more what this actually does? Or should that be a separate part of the docs?

wouterjdb (changes requested):
Some minor comments. Another (very) minor one: you could improve the consistency on interpunction (a dot or not at the end). 🖊️

src/flownet/config_parser/_config_parser.py
anders-kiaer: We can do that, the argument for writing Eclipse is that currently we sort of require Eclipse input case (Flow does not support writing `WSTATUS` which is used to instruct well status in FlowNet). 

Still change to `Simulation` @wouterjdb?

wouterjdb: Yeah, let's just change it.

wouterjdb (approved):
LGTM 🚀

PR #155, opened 2020-08-28
Kriging using well logs to constrain permeability and porosity parameter distributions in flow tubes

### LINKED ISSUE #154: Test: Include log data into FlowNet
Including log data to improve priors has been something we wanted to test out for a long time. Now that we are working with Norne we see that an improved prior might be needed.

It is therefore suggested to test out including log information (permeability and porosity) to steer the prior distributions of the FlowNet for these quantities.

A possible implementation would be to (from the simulation model as input) calculate the (harmonic?) mean permeability along the connections and assign an uncertainty. Then, assign distributions to all connections based on nearby measured values.

The following image is an attempt to visualize the envisioned methodology:

image

Any input @olelod @anders-kiaer @olwijn @edubarrosTNO @tayloris?

wouterjdb:
This pull request adds the following functionality:

- Loading of permeability and porosity logs from simulation data
- Kriging of measured data
- Assignment of value + uncertainty to individual tubes/connections based on kriging results.
- Many small formatting changes due to an update in black.

### Kriging example of log10(k) on Norne data
image

---

### Contributor checklist

- [x] :tada: This PR closes #154
   - [x] Load permeability and porosity data
   - [x] Perform kriging
   - [x] Assign permeability to tubes/connections
   - [x] Assign porosity to tubes/connections
  - [x] Perform tests, compare conditioned and non-conditioned prior ensemble for Norne case


src/flownet/ahm/_run_ahm.py
```diff
@@ -215,6 +216,99 @@ def _get_distribution(
...
+        anisotropy_scaling_z=anisotropy_scaling_z,
+    )
+
+    parameter_min_kriging = k3d3_interpolator(
+        network.connection_midpoints
+    ) - ss3d_interpolator(network.connection_midpoints)
+
+    parameter_max_kriging = k3d3_interpolator(
+        network.connection_midpoints
+    ) + ss3d_interpolator(network.connection_midpoints)
```
olelod, line 277: `ss3d_interpolator` gives the variance (sigma squared). should you not use the standard deviation to define the min/max (ref https://en.wikipedia.org/wiki/68%E2%80%9395%E2%80%9399.7_rule)?

wouterjdb: Good catch. Let's go for

,

which is the 95% confidence interval. Agreed?

olelod: I guess we can start there, at least. For porosity it should give a much broader range (since the variance is << 1).

wouterjdb: It might also explain why it worked _not so well_... 😄 😃

src/flownet/ahm/_run_ahm.py
```diff
@@ -215,6 +216,99 @@ def _get_distribution(
...
+        variogram_parameters = dict(
+            config.flownet.constraining.kriging.permeability_variogram_parameters
+        )
+    elif measurement_type == "porosity":
+        variogram_parameters = dict(
+            config.flownet.constraining.kriging.porosity_variogram_parameters
+        )
+
+    if not variogram_parameters:
+        variogram_parameters = None
```
anders-kiaer, line 260: When is this executing in this context? If `varioprogram_parameters` is an empty `{}`?

src/flownet/config_parser/_config_parser.py
```diff
@@ -209,6 +213,63 @@ def _to_abs_path(path: Optional[str]) -> str:
...
+                                    "enabled": {
+                                        MK.Type: types.Bool,
+                                        MK.Default: False,
+                                        MK.Description: "Switch to enable or disable kriging on well log data.",
+                                    },
+                                    "n": {
+                                        MK.Type: types.Integer,
+                                        MK.Default: 20,
+                                        MK.Description: "Number of kriged values in each direct. E.g, n = 10 -> "
+                                        "10x10x10 = 1000 values. Default: 20",
```
anders-kiaer, line 231: If default values are not automatically shown with `ConfigSuite` we should make a feature request there. If it is, or will be in near future, we could consider removing `Default: `?

src/flownet/utils/kriging.py
```diff
@@ -0,0 +1,66 @@
...
+) -> Tuple[RegularGridInterpolator, RegularGridInterpolator]:
+    """
+    Executes spatial kriging of input data and returns an scipy.interpolate.RegularInterpolator object.
+
+    Args:
+        data: 3xN np.ndarray with columns X, Y, Z, MEASUREMENT
+        n: Number of kriged values in each direct. E.g, n = 10 -> 10x10x10 = 1000 values
+        variogram_model: Variogram options as allowed in pykrige
+        nlags: nlag as defined in pykrige
+        anisotropy_scaling_<: anisotropy in the z-scale
```
anders-kiaer, line 25:
```suggestion
        anisotropy_scaling_z: anisotropy in the z-scale
```

anders-kiaer (changes requested):
Nice work :rocket: Some minor comments only.

src/flownet/ahm/_run_ahm.py
wouterjdb: Yes, so when the measurement type is set, but the variogram parameters are not set (that is, an empty dict - this is what configsuite spits out) then I actually want the None to be passed around and not an empty dict..

src/flownet/config_parser/_config_parser.py
wouterjdb: Default values are not shown in our current documentation. I agree with what you are saying. Will remove and make an issue.

wouterjdb: Two issues already existed on that, one a few weeks old: https://github.com/equinor/configsuite/issues/164

Guess they have it on the radar so we can drop the written defaults.

olelod (approved):
LGTM! Nice job!

.github/workflows/flownet.yml
```diff
@@ -46,10 +46,10 @@ jobs:
           INPUT_MODEL_FOLDER: ./flownet-testdata/${{ matrix.flownet-model }}/input_model
           # If you want the CI to (temporarily) run against your fork of the testdada,
           # change the value her from "equinor" to your username.
-          TESTDATA_REPO_OWNER: equinor
+          TESTDATA_REPO_OWNER: wouterjdb
           # If you want the CI to (temporarily) run against another branch than master,
           # change the value her from "master" to the relevant branch name.
-          TESTDATA_REPO_BRANCH: master
+          TESTDATA_REPO_BRANCH: add-disabled-kriging-to-norne
```
anders-kiaer, line 52: Change back

anders-kiaer (approved):

PR #199, opened 2020-09-22
Make FlowNet work with pyscal 0.6.x, and add possibility of using pyscal's SCALrecommendation 

### LINKED ISSUE #166: Running the Norne case with pyscal 0.6.0 results in a warning + numerical issues
The following warning is triggered:

https://github.com/equinor/pyscal/blob/0e5aaf0df49b47d229febeb41f778b817bf220ee/pyscal/wateroilgas.py#L254

All simulation runs fail with numerical issues.

From pyscal 0.6.0 the implementation of some endpoint have changed, see https://github.com/equinor/pyscal/releases/tag/v0.6.0.

Workaround: Running with pyscal 0.5.1 works fine.

### LINKED ISSUE #195: Interpolation between low/base/high cases in pyscal
With pyscal it is possible to interpolate between three user provided cases (low/base/high or pessimistic/base/optimistic) for relative permeability. We should check if we have something to gain by using this functionality - it would certainly reduce the number of parameters in the history matching.

olelod:
In pyscal 0.6.x the krowend/krogend parameters are deprecated. Only kroend should be used now (both kro(swl) and kro(sgcr) should equal kroend).

pyscal also has the possibility of taking three sets of parameter values as input (low/base/high), and making SWOF/SGOF tables that are interpolated between the low/base/high inputs. This would potentially reduce the number of parameters passed on to ERT for history matching a lot.

---

### Contributor checklist

- [x] :tada: This PR closes #166 and closes #195 .
   - [x] Remove deprecated parameters krowend/krogend from config_file. Adding kroend.
   - [x] Adding possibility of defining a base value for the relperm parameter, and use the SCALreccomendation functionality in pyscal to interpolate between the sets of curves


.github/workflows/flownet.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           TESTDATA_REPO_OWNER: equinor
           # If you want the CI to (temporarily) run against another branch than master,
           # change the value her from "master" to the relevant branch name.
-          TESTDATA_REPO_BRANCH: master
+          TESTDATA_REPO_BRANCH: i166-pyscal060
```
wouterjdb, line 52: Change before merging.

CHANGELOG.md
```diff
@@ -5,11 +5,14 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [Unreleased]
 
 ### Added
+- [#199](https://github.com/equinor/flownet/pull/199) Added the possibility of defining a 'base' value for all parameters related to relative permeability. This provides the opportunitty to interpolate between three relative permeability models. All the 'min' values will be used to construct a pessimistic or low model, the 'base' values are used for a 'base' model, and the 'high' values are used for an optimistic or high model value. The history matching can then be done with only one (two for three phase models - oil/water and gas/oil interpolated independently) relative permeability parameter(s) per SATNUM region.
```
wouterjdb, line 8:
```suggestion
- [#199](https://github.com/equinor/flownet/pull/199) Added the possibility of defining a 'base' value for all parameters related to relative permeability. This provides the opportunity to interpolate between three relative permeability models. All the 'min' values will be used to construct a pessimistic or low model, the 'base' values are used for a 'base' model, and the 'high' values are used for an optimistic or high model value. The history matching can then be done with only one (two for three phase models - oil/water and gas/oil interpolated independently) relative permeability parameter(s) per SATNUM region.
```

src/flownet/parameters/_relative_permeability.py
```diff
@@ -136,9 +134,9 @@ def __init__(
         satnum: pd.DataFrame,
         phases: List,
         fast_pyscal: bool = False,
+        interpolation_values: Optional[pd.DataFrame] = None,
```
wouterjdb, line 137: Update function documentation.

src/flownet/parameters/_relative_permeability.py
```diff
@@ -151,8 +149,17 @@ def __init__(
             distribution_values["parameter"].unique()
         )
         self._satnum: pd.DataFrame = satnum
-        self._swof, self._sgof = self._check_parameters()
         self._phases = phases
+        self._interpolation_values: Optional[pd.DataFrame] = None
+        self._scal_for_interp: Optional[PyscalList] = None
+        if isinstance(interpolation_values, pd.DataFrame):
```
wouterjdb, line 155: As `interpolation_values` is will be evaluated as `True` when it is a `pd.DataFrame` and `False` when it is `None`, you can drop the `isinstance()` call.

src/flownet/parameters/_relative_permeability.py
```diff
@@ -165,19 +172,23 @@ def _check_parameters(self) -> Tuple[bool, bool]:
             A tuple of booleans defining whether to generate the SWOF and or SGOF tables.
 
         """
-        # Check if the number of parameters make either sense for SWOF (8), SGOF (8) or both (14)
-        if len(self._parameters) != 8 and len(self._parameters) != 14:
+        if isinstance(self._interpolation_values, pd.DataFrame):
```
wouterjdb, line 175: As `interpolation_values` is will be evaluated as `True` when it is a `pd.DataFrame` and `False` when it is `None`, you can drop the `isinstance()` call.

src/flownet/parameters/_relative_permeability.py
```diff
@@ -260,30 +271,47 @@ def render_output(self) -> Dict:
...
-                    parameters, executor.map(partial_gen_wo, parameters)
-                ):
-                    str_swofs += relperm.SWOF(header=False)
-        elif self._sgof:
-            with concurrent.futures.ProcessPoolExecutor() as executor:
-                for i, relperm in zip(  # type: ignore[assignment]
-                    parameters, executor.map(partial_gen_og, parameters)
-                ):
-                    str_sgofs += relperm.SGOF(header=False)
+        if isinstance(self._interpolation_values, pd.DataFrame):
```
wouterjdb, line 274: As `interpolation_values` is will be evaluated as `True` when it is a `pd.DataFrame` and `False` when it is `None`, you can drop the `isinstance()` call.

src/flownet/parameters/_relative_permeability.py
```diff
@@ -260,30 +271,47 @@ def render_output(self) -> Dict:
...
+                    )
+                    if self._swof:
+                        str_swofs += relperm.SWOF(header=False)
+                    if self._sgof:
+                        str_sgofs += relperm.SGOF(header=False)
+                    if not self._swof and not self._sgof:
+                        raise ValueError(
+                            "It seems like both SWOF and SGOF should not be generated."
+                            "Either one of the two should be generated. Can't continue..."
+                        )
```
wouterjdb, line 289: One can use `enumerate()` for this:

```suggestion
            for i, parameter in enumerate(self._unique_satnums):
                if self._scal_for_interp is not None:
                    relperm = self._scal_for_interp[i + 1].interpolate(
                        parameter.get("interpolate"),
                        parameter.get("interpolate gas"),
                    )
                    if self._swof:
                        str_swofs += relperm.SWOF(header=False)
                    if self._sgof:
                        str_sgofs += relperm.SGOF(header=False)
                    if not self._swof and not self._sgof:
                        raise ValueError(
                            "It seems like both SWOF and SGOF should not be generated."
                            "Either one of the two should be generated. Can't continue..."
                        )
```

src/flownet/parameters/_relative_permeability.py
```diff
@@ -260,30 +271,47 @@ def render_output(self) -> Dict:
...
         else:
-            raise ValueError(
-                "It seems like both SWOF and SGOF should not be generated."
-                "Either one of the two should be generated. Can't continue..."
-            )
+            if self._swof and self._sgof:
+                with concurrent.futures.ProcessPoolExecutor() as executor:
+                    for i, relperm in zip(  # type: ignore[assignment]
```
wouterjdb, line 293: `i` isn't used, it is common practice to use an underscore in such case.

```suggestion
                    for _, relperm in zip(  # type: ignore[assignment]
```

src/flownet/parameters/_relative_permeability.py
```diff
@@ -260,30 +271,47 @@ def render_output(self) -> Dict:
...
+                        parameters, executor.map(partial_gen_wog, parameters)
+                    ):
+                        str_swofs += relperm.SWOF(header=False)
+                        str_sgofs += relperm.SGOF(header=False)
+            elif self._swof:
+                with concurrent.futures.ProcessPoolExecutor() as executor:
+                    for i, relperm in zip(  # type: ignore[assignment]
```
wouterjdb, line 300: Underscore.

src/flownet/parameters/_relative_permeability.py
```diff
@@ -260,30 +271,47 @@ def render_output(self) -> Dict:
...
+                        parameters, executor.map(partial_gen_wo, parameters)
+                    ):
+                        str_swofs += relperm.SWOF(header=False)
+            elif self._sgof:
+                with concurrent.futures.ProcessPoolExecutor() as executor:
+                    for i, relperm in zip(  # type: ignore[assignment]
```
wouterjdb, line 306: Underscore.

src/flownet/parameters/_relative_permeability.py
```diff
@@ -260,30 +271,47 @@ def render_output(self) -> Dict:
...
-                ):
-                    str_swofs += relperm.SWOF(header=False)
-        elif self._sgof:
-            with concurrent.futures.ProcessPoolExecutor() as executor:
-                for i, relperm in zip(  # type: ignore[assignment]
-                    parameters, executor.map(partial_gen_og, parameters)
-                ):
-                    str_sgofs += relperm.SGOF(header=False)
+        if isinstance(self._interpolation_values, pd.DataFrame):
+            for i in range(len(self._unique_satnums)):
```
wouterjdb, line 275: Use `enumerate()`, see comment below.

wouterjdb (changes requested):
Looks good - some minor things only.

src/flownet/parameters/_relative_permeability.py
olelod: If i drop the isinstance (which was what i did at first), i get 'ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().'

wouterjdb: OK - then ignore my comment - it's wrong in this case then.

src/flownet/parameters/_relative_permeability.py
olelod: Enumerate over parameters then (even starting at index "1"). Looks better, thanks.

wouterjdb (approved):
LGTM 🚀 

There are some missing descriptions in the config parser setup, but that was already the case and outside of the scope of this PR. I've made an issue on this #204.

PR #205, opened 2020-10-08
Add python 3.8 to github workflow

wouterjdb:
Add Python 3.8 to github workflows.

---

### Contributor checklist

   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/src/flownet/ahm/_run_ahm.py
+++ b/src/flownet/ahm/_run_ahm.py
@@ -389,7 +389,7 @@ def run_flownet_history_matching(
     relperm_parameters = config.model_parameters.relative_permeability.regions[
         0
     ]._asdict()
-    relperm_parameters.popitem(last=False)
+    relperm_parameters.popitem(False)
 
     relperm_dict = {
         key: values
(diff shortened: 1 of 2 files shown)


src/flownet/ahm/_run_ahm.py
```diff
@@ -389,7 +389,7 @@ def run_flownet_history_matching(
     relperm_parameters = config.model_parameters.relative_permeability.regions[
         0
     ]._asdict()
-    relperm_parameters.popitem(last=False)
+    relperm_parameters.popitem()
```
wouterjdb, line 392: @olelod 

For some reason only in Python 3.8 I'm getting complaints about the argument that should not be there. I then bumped into this:

> Definition and Usage
> The popitem() method removes the item that was last inserted into the dictionary. In versions before 3.7, the > popitem() method removes a random item.
> 
> The removed item is the return value of the popitem() method, as a tuple, see example below.

Could this leads to differences in execution for various versions? Or is there only 1 item when you do this?

olelod: I'm just removing the 'id' key. I guess it's maybe better to use `relperm_parameters.pop('id',None)` instead?

PR #216, opened 2020-10-14
Refactor SUMMARY.inc and add vectors

### LINKED ISSUE #211: Additional summary vectors for the Norne case
To gain understanding on the matching of Norne, it would be useful to add a few more summary vectors in the reference Eclipse and FlowNet simulations. A few vectors would be:
- FLPT, FLPR (to track liquid production)
- FVPT, FVPR (reservoir volume production)
- FPR (average reservoir pressure)
- FMWPR (number of open producers)
- FMWIN (number of open injectors)
- WMCON (number of open well connections)

To be checked: if OPM-Flow supports all of these.

wouterjdb:
Refactor the `SUMMARY.inc` static file and add requested vectors.

---

### Contributor checklist

- [x] :tada: This PR closes #211.
   - [x] Refactor the static file
   - [x] Add vectors as request in issue

### DIFF
--- a/src/flownet/static/SUMMARY.inc
+++ b/src/flownet/static/SUMMARY.inc
@@ -101,7 +101,6 @@ WWPRH
 /
 
 -- Production cumulatives
-
 WOPT
 /
 WGPT
--- a/src/flownet/static/SUMMARY.inc
+++ b/src/flownet/static/SUMMARY.inc
@@ -12,7 +12,6 @@ FOPRH
 FGPRH
 FWPRH
 FLPRH
-FVPRH
 
 -- Production cumulatives
 FOPT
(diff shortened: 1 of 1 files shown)


src/flownet/static/SUMMARY.inc
```diff
@@ -1,105 +1,155 @@
...
+FLPR
+FVPR
 FOPRH
-FOPTH
-FWPT
-FWPTH
+FGPRH
+FWPRH
+FLPRH
+FVPRH
```
olelod, line 15: Quickly scanning through the tables in the OPM manual, and it seems like FVPRH is not supported. Other than that it looks all good.

olelod (approved):

PR #220, opened 2020-10-22
Adding possibility to change control modes in history matching

### LINKED ISSUE #219: Enable HM controls in the FlowNet config file
It could be useful to enable HM controls / targets for producers (WCONHIST) and injectors (WCONINJH) to be set in the FlowNet config file. Right now, the target 'RESV' is set for producers (where a few other options would be possible) and 'RATE' for injectors (where it could be also 'BHP').

olelod:
Adding the possibility to change how the production wells are controlled in the WCONHIST keywords. This will not be controlled by one keyword in the config yaml.

---

### Contributor checklist

- [x] :tada: This PR closes #219 
   - [x] Setting keyword for control mode in WCONHIST and WCONINJH in the config yaml

### DIFF
--- a/src/flownet/realization/_schedule.py
+++ b/src/flownet/realization/_schedule.py
@@ -27,7 +27,8 @@ class Schedule:
         self._schedule_items: List = []
         self._network: NetworkModel = network
         self._df_production_data: pd.DataFrame = df_production_data
-        self._control_mode: str = config.flownet.prod_well_control_mode
+        self._prod_control_mode: str = config.flownet.prod_control_mode
+        self._inj_control_mode: str = config.flownet.inj_control_mode
         self._case_name: str = config.name
         self._create_schedule()
 
@@ -161,7 +162,7 @@ class Schedule:
                         date=value["date"],
                         well_name=value["WELL_NAME"],
                         status=value["WSTAT"],
-                        control_mode=self._control_mode,
+                        control_mode=self._prod_control_mode,
                         vfp_table=vfp_tables[value["WELL_NAME"]],
                         oil_rate=value["WOPR"],
                         water_rate=value["WWPR"],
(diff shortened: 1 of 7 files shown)


src/flownet/realization/_schedule.py
```diff
@@ -197,6 +197,7 @@ def _calculate_wconinjh(self):
                         rate=value["WWIR"],
                         bhp=value["WBHP"],
                         thp=value["WTHP"],
+                        target=self._inj_control_mode,
```
wouterjdb, line 200: And then this `inj_control_mode`?

src/flownet/realization/_schedule.py
```diff
@@ -494,35 +496,30 @@ def get_nr_observations(self, training_set_fraction: float) -> int:
         num_training_dates = round(len(self.get_dates()) * training_set_fraction)
         i = 0
 
-        for date in self.get_dates()[0 : num_training_dates - 1]:
+        for date in self.get_dates()[1 : num_training_dates - 1]:
```
wouterjdb, line 499: Why should this be 1? Now you are skipping the first date?

wouterjdb (changes requested):

src/flownet/realization/_schedule.py
olelod: For some reason the code used to remove the first date for all wells. I removed that, but then ERT gives an error about the first date of the simulation: `** ERROR: It is unfortunately not possible to use summary observations from the
          start of the simulation. Problem with observation:WGPR_B-2H_06_11_1997 at 06/11/1997`. Then i just removed the first date (both in generating the `observations.ertobs` and in counting the number of observations).

src/flownet/realization/_schedule.py
```diff
@@ -162,7 +162,7 @@ def _calculate_wconhist(self):
                         date=value["date"],
                         well_name=value["WELL_NAME"],
                         status=value["WSTAT"],
-                        control_mode=self._control_mode,
+                        control_mode=self._prod_control_mode,
```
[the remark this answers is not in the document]
olelod: I just used the ones allready defined, but i agree with your comment about clarity.

src/flownet/realization/_schedule.py
olelod: Same here

wouterjdb (approved):
LGTM 🚀

PR #227, opened 2020-10-30
Add export of training, test, and complete observation files

### LINKED ISSUE #225: Make it possible to have training, test, and complete observation files
Currently the ertobs file is only containing the training observations, the yaml file always contains everything.

I suggest we, by default, create for both formats a training, test and complete file.

I also suggest to make a python test which checks that the data in both outputted formats is identical, as we now have had multiple times issues with this (see for example #222).

So the tasks:

- [ ] Create a train, test and complete ertobs and yaml file by default
- [ ] Create a test which verifies data written to a yaml and ertobs file is identifical

tayloris:
Now, both yaml and ert observation file has the same information, and both ignore the first simulation day. Aditional test and training observation files are created and the corresponding test check if the has the same information.

---

### Contributor checklist

- [x] :tada: This PR closes #225 
   - [x] Create a train, test and complete ertobs and yaml file by default
   - [x] Create a test which verifies data written to a yaml and ertobs file is identifical
- [x] :robot: I have added tests, that could be use in issue 227.

After revision
- [x] Generate some dummy production data (or have a testdata set)
- [x] Read the data and  write all 6 observation files
    - [x] Read the data
    - [x] write all 6 observation files using your modified code (train, test, complete for yaml and ertobs)
- [x] Read the newly created files and compare if they are pairwise eq


tayloris:
I think there could be better (faster) ways to compare dictionaries and files in the test. But I'm not fully aware of functionalities that could improve it.

tayloris:
Additionally, the template for yaml observation file could be written in a more elegant way, but it works as it is now.

src/flownet/ert/_create_ert_setup.py
```diff
@@ -47,28 +47,94 @@ def _create_observation_file(
         Nothing
 
     """
-    num_training_dates = round(len(schedule.get_dates()) * training_set_fraction)
+    num_dates = len(schedule.get_dates())
+    num_training_dates = round(num_dates * training_set_fraction)
 
     if yaml:
+        observations_yamlobs = obs_file
```
wouterjdb, line 54: When creating the three files you repeat a lot of code. You could consider using a for loop, maybe just a single one, that gets a list of settings you want to export an observation file for.

Like so (untested code, indentation is messy):
```python


export_settings = [ 
                             ["complete", 1, num_dates],
                             ["training", 1, num_training_dates],
                             ["test", num_training_dates + 1, num_training_dates + 1],
                             ["complete", 1, num_dates],
                             ["training", 1, num_training_dates],
                             ["test", num_training_dates + 1, num_dates],
                        ]

prefix_name = os.path.splitext(obs_file)

for _, setting in enumerate(export_settings):
    if yaml:
        obs_export_type = "yaml"
    else:
        obs_export_type = "ertobs"
    export_filename = f"{prefix_name}_{setting[0]}.{obs_export_type}"
    template = _TEMPLATE_ENVIRONMENT.get_template(f"observations.{obs_export_type}.jinja2")
   
    with open(export_filename , "w") as fh:
            fh.write(
                template.render(
                    {
                        "schedule": schedule,
                        "error_config": config.flownet.data_source.simulation.vectors,
                        "num_beginning_date": setting[1],
                        "num_end_date": setting[2]
                    }
                )
            )
```

This potentially replaces lines 53-137. It would however require that there is only 1 template for ertobs, and 1 template for yaml obs. Which is directly another point I'd like to raise: by increasing the number of template the update job for a developer becomes even worse compared to updating two files, as we have to do per today. If there is no good argument for have 6 templates for 2 different files, I'd suggest to drop 4 of them.

src/flownet/ert/_create_ert_setup.py
```diff
@@ -230,17 +296,25 @@ def create_ert_setup(  # pylint: disable=too-many-arguments
...
+    ert_output_file_names = os.path.join(output_folder, "observations.ertobs")
+
+    yaml_output_file_names = os.path.join(output_folder, "observations.yamlobs")
+
     if not prediction_setup:
         if parameters is not None:
             _create_observation_file(
                 schedule,
-                output_folder / "observations.ertobs",
+                ert_output_file_names,
```
wouterjdb, line 307: If you revert back to `pathlib` you can do this type of operation again, which the standard/preferred way of doing operations with paths used in FlowNet.

src/flownet/templates/observations.yamlobs.jinja2
```diff
@@ -5,7 +5,8 @@ smry:
    {% if loop.first and not loop.last: %}
  - key: WOPR:{{ well_name }}
    observations:
-   {%- elif not loop.first: %}
+   {%- endif -%} 
+   {%- if (kw.date.toordinal() > schedule.get_first_date().toordinal()) and (kw.date.toordinal() >= schedule.get_dates()[num_beginning_date-1].toordinal() )  and (kw.date.toordinal() <= schedule.get_dates()[num_end_date-1].toordinal()) %}
```
wouterjdb, line 9: Converting everything to ordinals is a bit messy. Can't you just keep working with pydates?

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
+from datetime import datetime
+
+import os.path
+
+# import pytest
```
wouterjdb, line 5: Remove if not in use.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
+from datetime import datetime
+
```
wouterjdb, line 2: Remove blank line.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
...
+import os.path
+
+# import pytest
+
+import yaml
+
+
+def read_ert_obs(ert_obs_file_name):
```
wouterjdb, line 10:
```suggestion
def read_ert_obs(ert_obs_file_name: str) -> dict:
```

However, maybe even nicer would be to use pathlib for this as well?

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
...
+    # Reading ERT file
+    assert os.path.exists(ert_obs_file_name) == 1
+    a_ert_file = open(ert_obs_file_name, "r")
```
wouterjdb, line 13: Always use `with` to open a file.

Like so:

```python
with open(xyz,"r") a fh:
    for line in .......
```
When you leave the with block you automatically close the file.

Don't forget documentation.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
...
+            ert_obs[dic["KEY"]][0].append(
+                datetime.strptime(dic["DATE"], "%d/%m/%Y").toordinal()
+            )
+            ert_obs[dic["KEY"]][1].append(float(dic["VALUE"]))
+            ert_obs[dic["KEY"]][2].append(float(dic["ERROR"]))
+
+    return ert_obs
+
+
+def read_yaml_obs(yaml_obs_file_name):
```
wouterjdb, line 42: Use type annotations. Also for the reutnr type.

Add documentation.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
...
+    # Reading YALM file
+    assert os.path.exists(yaml_obs_file_name) == 1
+    a_yaml_file = open(yaml_obs_file_name, "r")
+    yaml.allow_duplicate_keys = True
+
+    return yaml.load(a_yaml_file, Loader=yaml.FullLoader)
+
+
+def compare(ert_obs_dict, yaml_obs_dict):
```
wouterjdb, line 51: Typing. Docs.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
...
+            for lost_item in list_item["observations"]:
+                if not list_item["key"] in yaml_obs:
+                    yaml_obs[list_item["key"]] = [[], [], []]
+                yaml_obs[list_item["key"]][0].append(lost_item["date"].toordinal())
+                yaml_obs[list_item["key"]][1].append(float(lost_item["value"]))
+                yaml_obs[list_item["key"]][2].append(float(lost_item["error"]))
+            assert yaml_obs[list_item["key"]][0] == ert_obs_dict[list_item["key"]][0]
+            assert yaml_obs[list_item["key"]][1] == ert_obs_dict[list_item["key"]][1]
+            assert yaml_obs[list_item["key"]][2] == ert_obs_dict[list_item["key"]][2]
+            return equal
```
wouterjdb, line 65: This looks strange. Your return value will be shot back after the first iteration of the first nested loop?

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def test_check_obsfiles_ert_yaml():
```
wouterjdb, line 68: Typings. Docs.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
...
+    ert_obs_file_name = "./tests/observation_files/observations.ertobs"
```
wouterjdb, line 69: You could consider writing hard-coded paths as CAPITALIZED variables at the beginning of the file; this is a common practice.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
...
+    test_yaml_obs_file_name = "./tests/observation_files/observations_test.yamlobs"
+
+    # Comparing the complete observation data
+    # Reading ERT file
+    ert_obs = read_ert_obs(ert_obs_file_name)
+
+    # Reading YAML file
+    parsed_yaml_file = read_yaml_obs(yaml_obs_file_name)
+
+    assert compare(ert_obs, parsed_yaml_file) is True
```
wouterjdb, line 87: No need for `is True`

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # Comparing the training observation data
+    # Reading ERT file
+    ert_obs = read_ert_obs(training_obs_file_name)
+
+    # Reading YAML file
+    parsed_yaml_file = read_yaml_obs(training_yaml_obs_file_name)
+
+    assert compare(ert_obs, parsed_yaml_file) is True
```
wouterjdb, line 96: Remove `is True`

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # Comparing the Test observation data
+    # Reading ERT file
+    ert_obs = read_ert_obs(test_ert_obs_file_name)
+
+    # Reading YAML file
+    parsed_yaml_file = read_yaml_obs(test_yaml_obs_file_name)
+
+    assert compare(ert_obs, parsed_yaml_file) is True
```
wouterjdb, line 105: Remove `is True`

wouterjdb (changes requested):
Great start. I have added some comments to the PR, my main points in the review are:

- Try to reduce the 6 templates to 2 template. Or argue for why this cannot be done without 6 templates.
- Try to avoid repeating code (`src/flownet/ert/_create_ert_setup.py`). I wrote a suggestion of how this could be done. This is untested code, use it as "inspiration" only.
- Some missing typing and documentation / style issues.

I really like that we now have a test for this - that is going to come in handy!

src/flownet/templates/observations.yamlobs.jinja2
tayloris: I didn't find a way to compare "<=" or "<="between pydates. It appear that it just support < , > and ==. I'll read about it and find a better way.

tests/test_check_obsfiles_ert_yaml.py
tayloris: Yes, I intendent that. I got some troubles with pylint, I'll try to what was the problem

src/flownet/templates/observations.yamlobs.jinja2
wouterjdb: ?

image

tayloris:
It is ready for review now. I try to modify `_config_parser.py` but CI didn't let me. I just wanted to change  line 444 
 from `observations.yamlobs`  to `observations_complete.yamlobs`.

src/flownet/ert/_create_ert_setup.py
```diff
@@ -47,28 +47,34 @@ def _create_observation_file(
...
-    else:
-        template = _TEMPLATE_ENVIRONMENT.get_template("observations.ertobs.jinja2")
-        with open(obs_file, "w") as fh:
+    export_settings = [
+        ["_complete", 1, num_dates],
+        ["_training", 1, num_training_dates],
+        ["_test", num_training_dates + 1, num_dates],
+    ]
+
+    prefix_name = os.path.splitext(obs_file)
```
wouterjdb, line 59: The `splitext()` function returns a tuple `(root, ext)`. You are only interested in the root of the file, so take out that part right away:

```suggestion
    file_root = os.path.splitext(obs_file)[0]
```

src/flownet/ert/_create_ert_setup.py
```diff
@@ -47,28 +47,34 @@ def _create_observation_file(
...
+
+    for _, setting in enumerate(export_settings):
+        if yaml:
+            obs_export_type = "yamlobs"
+        else:
+            obs_export_type = "ertobs"
+        export_filename = f"{prefix_name[0]}{setting[0]}.{obs_export_type}"
```
wouterjdb, line 66:
```suggestion
        export_filename = f"{file_root}{setting[0]}.{obs_export_type}"
```

src/flownet/ert/_create_ert_setup.py
```diff
@@ -47,28 +47,34 @@ def _create_observation_file(
...
-        with open(obs_file, "w") as fh:
+    export_settings = [
+        ["_complete", 1, num_dates],
+        ["_training", 1, num_training_dates],
+        ["_test", num_training_dates + 1, num_dates],
+    ]
+
+    prefix_name = os.path.splitext(obs_file)
+
+    for _, setting in enumerate(export_settings):
```
wouterjdb, line 61: Because the first item of the tuple from enumerate isn't used; this should work as well:

```suggestion
    for setting in export_settings:
```

src/flownet/realization/_schedule.py
```diff
@@ -315,7 +315,7 @@ def get_first_date(self) -> datetime.date:
 
     def get_keywords(
         self,
-        date: datetime.date = None,
+        date: Optional[Union[List[datetime.date], datetime.date]] = None,
```
wouterjdb, line 318: Maybe change to "dates" to reflect it is not just only 1?
```suggestion
        dates: Optional[Union[List[datetime.date], datetime.date]] = None,
```

src/flownet/realization/_schedule.py
```diff
@@ -325,7 +325,7 @@ def get_keywords(
         particular keyword class.
 
         Args:
-            date: Date at which to lookup keywords
+            date: Date or list of dates at which to lookup keywords
```
wouterjdb, line 328: Update docstring to "dates". Also in the descriptiona few lines above it should be changed.

"at a given date" isn't correct anymore.

src/flownet/templates/ahm_config.ert.jinja2
```diff
@@ -37,6 +37,6 @@ GEN_KW FLOWNET_PARAMETERS ./EMPTYFILE ./EMPTYFILE ./parameters.ertparam
 SIMULATION_JOB DELETE_IN_CURRENT_ITERATION <ECLBASE>
 {%- endif %}
 
-OBS_CONFIG ./observations.ertobs
+OBS_CONFIG ./observations_complete.ertobs
```
wouterjdb, line 40: This should be the `observations_training.ertobs` file.

src/flownet/templates/observations.yamlobs.jinja2
```diff
@@ -1,10 +1,13 @@
...
  {%- for well_name in schedule.get_wells(kw_class="WCONHIST") -%}
    {%- if (error_config.WOPR.rel_error is not none) and (error_config.WOPR.min_error is not none) -%}
-   {%- for kw in schedule.get_keywords(kw_class="WCONHIST", well_name=well_name, ignore_nan="oil_rate") -%}
+   {%- for kw in schedule.get_keywords(kw_class="WCONHIST", well_name=well_name, ignore_nan="oil_rate", date=schedule.get_dates()[num_beginning_date:num_end_date]) -%}
    {% if loop.first and not loop.last: %}
  - key: WOPR:{{ well_name }}
    observations:
+    - date: {{ kw.date.strftime('%Y-%m-%d') }}
+      value: {{ kw.oil_rate }}
+      error: {{ [error_config.WOPR.rel_error * kw.oil_rate, error_config.WOPR.min_error] | max }}
```
wouterjdb, line 10:
```suggestion
    - date: {{ kw.date.strftime('%Y-%m-%d') }}
      value: {{ kw.oil_rate }}
      error: {{ [error_config.WOPR.rel_error * kw.oil_rate, error_config.WOPR.min_error] | max }}
```

Isn't this double (see line below). I think you could end the if statement just behind the `observations:`. Then you don't have to write the date/value/error twice the same.

This exact same reduction can potentially be applied 9 times which saves quite a few lines of code.

src/flownet/templates/webviz_ahm_config.yml.jinja2
```diff
@@ -17,7 +17,7 @@ pages:
         {%- for iteration in [iterations[0], iterations[-1]] %}
          - iter-{{ iteration }}
         {%- endfor %}
-        observation_file: ./observations.yamlobs
+        observation_file: ./observations_complete.yamlobs
```
wouterjdb, line 20: These observations will be used in the misfit calculation and that should only be calculated on the training observations.

```suggestion
        observation_file: ./observations_training.yamlobs
```

src/flownet/templates/webviz_ahm_config.yml.jinja2
```diff
@@ -32,7 +32,7 @@ pages:
         {%- for iteration in [iterations[0], iterations[-1]] %}
          - iter-{{ iteration }}
         {%- endfor %}
-        obsfile: ./observations.yamlobs
+        obsfile: ./observations_complete.yamlobs
```
wouterjdb, line 35: @anders-kiaer Is it possible to supply two observation file for the summary plotting? It would be nice to supply both the training and the test set and have them colored in a different way.

wouterjdb (changes requested):
Looks good, some minor comments in the code.

I'm a bit confused what you are actually testing? To me it looks like you are testing whether or not you can read existing (test data) observation files and whether they are equal.

This doesn't actually test the code you implemented, which is about writing the observation files. I'd say that what you could do is:

1. Generate some dummy production data (or have a testdata set)
2. Read the data, write all 6 observation files using your modified code (train, test, complete for yaml and ertobs)
3. Read the newly created files and compare if they are pairwise equal.

src/flownet/templates/webviz_ahm_config.yml.jinja2
wouterjdb: Anders informed me that different colors, or multiple observation files, are not possible. However, you could add an extra yaml key called `comment` and write down test the observation belongs to train or test.

Like so:

```
    observations:
      - {value: 251, error: 10, date: 2003-01-01, comment: First measurement after sensor drift correction}
```

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,259 @@
...
+           None. The file names containig the observation files are harcoded as
+       Returns:
+           None
+    """
+
+###############################################################################################                   
+###############################################################################################                   
+
+    # Load Config
+    config = parse_config(_CONFIG_FILE_NAME, None)
```
tayloris, line 127: How Could I create a fake config variables.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,259 @@
...
+                    status=value["WSTAT"],
+                    prod_control_mode=schedule._prod_control_mode,
+                    vfp_table=vfp_tables[value["WELL_NAME"]],
+                    oil_rate=value["WOPR"],
+                    water_rate=value["WWPR"],
+                    gas_rate=value["WGPR"],
+                    bhp=value["WBHP"],
+                    thp=value["WTHP"],
+                )
+            )
```
tayloris, line 208: Remormulate Schedules functions: I copied and pasted internal lines in _calculate_wconhist() and `_calculate_wconinjh()`.  This functions requires `well_names` and `start_dates` already asing by _calculate_compdat()     `_calculate_welspecs() `. So I include `well_names` and `start_dates` hard coded. But this is not the solution. We need to reformulate this part.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,259 @@
...
+                    template.render(
+                        {
+                            "schedule": schedule,
+                            "error_config": config.flownet.data_source.simulation.vectors,
+                            "num_beginning_date": setting[1],
+                            "num_end_date": setting[2],
+                        }
+                    )
+                )
+
```
tayloris, line 245: This lines are meant to replicate  _create_observation_file. However,   I wasn't able to import  _create_observation_file  from flownet.ert  . It should be possible but I don't know how to do it.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,259 @@
...
+    # Load Config
+    config = parse_config(_CONFIG_FILE_NAME, None)
+    
+    #TODO: Create a fake error vector (config.flownet.data_source.simulation.vectors),
+    #      it apear that it needs to be named_dict type.
+    #vectors = configsuite.config.named_dict
+    #print(type(config.flownet.data_source.simulation.vectors))
+    #print(config.flownet.data_source.simulation.vectors)
+    #print(type(vectors))
+    #print(vectors)
```
tayloris, line 135: How Could I create a fake config variable? actually, what its needed is config.flownet.data_source.simulation.vectors and this is a name_dict variable, I don't know how to defined and assign values on it. I try to follow what is done in  Config_Parse but It didn't work.

wouterjdb: It's done in this file: https://github.com/equinor/flownet/blob/master/tests/test_generate_connections.py

wouterjdb:
Guess it is possible now to remove the many ertobs/yamlobs files that are in this PR?

tests/test_check_obsfiles_ert_yaml.py
wouterjdb: Just had chat with you :) Should be solved.

src/flownet/templates/webviz_ahm_config.yml.jinja2
tayloris: So do you want one observation yaml file that has inside all data and a tag for each observation that indicates if corresponds to test or training?

wouterjdb: Yes - but also still the training and test files. So the goal of the PR doesn't change.

src/flownet/realization/_schedule.py
```diff
@@ -30,20 +30,20 @@ def __init__(
         config: ConfigSuite.snapshot = None,
     ):
         self._schedule_items: List = []
-        self._prod_control_mode: str
-        self._inj_control_mode: str
+        self.prod_control_mode: str
```
wouterjdb, line 33: No need to remove the underscore. It is only for internal use still isn't it?

src/flownet/templates/webviz_ahm_config.yml.jinja2
```diff
@@ -32,7 +32,7 @@ pages:
         {%- for iteration in [iterations[0], iterations[-1]] %}
          - iter-{{ iteration }}
         {%- endfor %}
-        obsfile: ./observations.yamlobs
+        obsfile: ./observations_training.yamlobs
```
wouterjdb, line 35: The time series plots should be shown with the complete observations file (this is where the labels will be shown)

```suggestion
        obsfile: ./observations_complete.yamlobs
```

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,283 @@
+import pathlib
```
wouterjdb, line 1: Don't add enter between every import. Group them like

```
python internal package 1
python internal package 2

Third party package 1
Third party package 2

Flownet import 1
Flownet import 2
```

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,283 @@
...
+
+from flownet.realization import Schedule
+
+from flownet.ert import create_observation_file
+
+from flownet.realization._simulation_keywords import WCONHIST, WCONINJH
+
+
+_OBSERVATION_FILES = pathlib.Path("./tests/observation_files")
+
```
wouterjdb, line 25: No need for additional enters everywhere. Just make logical groupings.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,283 @@
...
+_TEMPLATE_ENVIRONMENT = jinja2.Environment(
+    loader=jinja2.PackageLoader("flownet", "templates"),
+    undefined=jinja2.StrictUndefined,
+)
+_TEMPLATE_ENVIRONMENT.globals["isnan"] = np.isnan
+
+
+def read_ert_obs(ert_obs_file_name: pathlib.Path) -> dict:
+    """This function reads the content of a ERT observation file and returns the information in a dictionary.
+    Args:
```
wouterjdb, line 43:
```suggestion

    Args:
```

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,283 @@
...
+            )
+            ert_obs[dic["KEY"]][1].append(float(dic["VALUE"]))
+            ert_obs[dic["KEY"]][2].append(float(dic["ERROR"]))
+
+    return ert_obs
+
+
+def read_yaml_obs(yaml_obs_file_name: pathlib.Path) -> dict:
+    """This function reads the content of a YAML observation file and returns the information in a dictionary.
+    Args:
```
wouterjdb, line 78:
```suggestion

    Args:
```

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,283 @@
...
+    assert os.path.exists(yaml_obs_file_name) == 1
+    a_yaml_file = open(yaml_obs_file_name, "r")
+    yaml.allow_duplicate_keys = True
+
+    return yaml.load(a_yaml_file, Loader=yaml.FullLoader)
+
+
+def compare(ert_obs_dict: dict, yaml_obs_dict: dict) -> bool:
+    """This function compares if the given dictionaries: ert_obs_dict and yaml_obs_dict contain the same information.
+    Args:
```
wouterjdb, line 92:
```suggestion

    Args:
```

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,283 @@
...
+                if not list_item["key"] in yaml_obs:
+                    yaml_obs[list_item["key"]] = [[], [], []]
+                yaml_obs[list_item["key"]][0].append(lost_item["date"].toordinal())
+                yaml_obs[list_item["key"]][1].append(float(lost_item["value"]))
+                yaml_obs[list_item["key"]][2].append(float(lost_item["error"]))
+            assert yaml_obs[list_item["key"]][0] == ert_obs_dict[list_item["key"]][0]
+            assert yaml_obs[list_item["key"]][1] == ert_obs_dict[list_item["key"]][1]
+            assert yaml_obs[list_item["key"]][2] == ert_obs_dict[list_item["key"]][2]
+
+    return True
```
wouterjdb, line 112: Do you actually need to return `True` at all? The code would already fail with the assert statements.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+def test_check_obsfiles_ert_yaml() -> None:
+    """
+    This function checks if the observation files (complete, training, and test) in ERT and YAML version are equal.
+       Args:
+           None. The file names containig the observation files are harcoded as
+       Returns:
+           None
+    """
```
wouterjdb, line 122: You function doesn't return `None`, neither does it take as an argument `None`. I suggest writing it like this:
```suggestion
    This function checks if the observation files (complete, training, and test) in ERT and YAML version are equal.

    Returns:
        Nothing
    """
```

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,283 @@
...
+                    vfp_table="1*",
+                    oil_rate=value["WOPR"],
+                    water_rate=value["WWPR"],
+                    gas_rate=value["WGPR"],
+                    bhp=value["WBHP"],
+                    thp=value["WTHP"],
+                )
+            )
+
+    print("Writing observations files")
```
wouterjdb, line 246: Remove print statement

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,283 @@
...
+
+    create_observation_file(
+        schedule,
+        _OBSERVATION_FILES / "observations.yamlobs",
+        config,
+        _TRAINING_SET_FRACTION,
+        yaml=True,
+    )
+
+    print("Reading observations files")
```
wouterjdb, line 262: Remove.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,283 @@
...
+    config.flownet.data_source.simulation.vectors.WGIR.min_error = _MIN_ERROR
+    config.flownet.data_source.simulation.vectors.WGIR.rel_error = _REL_ERROR
+
+    config.flownet.data_source.simulation.vectors.WWIR = collections.namedtuple(
+        "WWIR", "min_error"
+    )
+    config.flownet.data_source.simulation.vectors.WWIR.min_error = _MIN_ERROR
+    config.flownet.data_source.simulation.vectors.WWIR.rel_error = _REL_ERROR
+    # Load production
+    print("Reading production data")
```
wouterjdb, line 175: Remove.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -0,0 +1,283 @@
...
+from flownet.realization._simulation_keywords import WCONHIST, WCONINJH
+
+
+_OBSERVATION_FILES = pathlib.Path("./tests/observation_files")
+
+_PRODUCTION_DATA_FILE_NAME = pathlib.Path(_OBSERVATION_FILES / "ProductionData.csv")
+
+_TRAINING_SET_FRACTION = 0.75
+
+# Define a fake Config
```
wouterjdb, line 30: ?

wouterjdb (changes requested):
Looks much better. Some minor code style things left.

tests/test_check_obsfiles_ert_yaml.py
tayloris: We need to return true in the test, because outside of the function there is an assert, checking we say True

tayloris: We dont neet it at all

wouterjdb (approved):
LGTM 🚀

PR #249, opened 2020-11-13
Makes analysis workflow robust against failed simulations

### LINKED ISSUE #248: Analytics failing for runs with incomplete realizations / runs
image

edubarrosTNO:
Makes analysis workflow robust against incomplete (shorter than end date of analysis period) and failed simulations among the ensemble of FlowNet realizations

---

### Contributor checklist

- [x] :tada: This PR closes #248.
   - [x] Checks against absent summary files
   - [x] Checks against simulations shorter than end date of analysis period


src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -166,7 +165,7 @@ def accuracy_metric(data_reference, data_test, metric):
     if metric in ("MSE", "RMSE", "NRMSE"):
         score = mean_squared_error(data_reference, data_test)
         if metric in ("RMSE", "NRMSE"):
-            score = math.sqrt(score)
+            score = mean_squared_error(data_reference, data_test, squared=False)
```
wouterjdb, line 168: I'd suggest to make this function more readable by rewriting it to something like (lines 168-180):

```python
if metric is "MSE":
    score = mean_squared_error(data_reference, data_test)
elif metric is "RMSE":
    score = mean_squared_error(data_reference, data_test, squared=False)
elif metric is "NMRSE":
    score = mean_squared_error(data_reference, data_test, squared=False) / (np.amax(data_reference) - np.amin(data_reference))
elif metric is "MAE":
    score = mean_absolute_error(data_reference, data_test)
elif metric is "NMAE":
    score = mean_absolute_error(data_reference, data_test) / (np.amax(data_reference) - np.amin(data_reference))
elif metric is "R2":
    score = r2_score(data_reference, data_test, multioutput="variance_weighted")
else:
    raise NotImplementedError(f"The metric '{metric}' is not supported.")

return score
```

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -190,8 +189,9 @@ def _load_simulations(runpath: str, ecl_base: str) -> EclSum:
     Returns:
         EclSum
     """
-
-    return EclSum(str(pathlib.Path(runpath) / pathlib.Path(ecl_base)))
+    if os.path.exists(pathlib.Path(runpath) / pathlib.Path(ecl_base + ".UNSMRY")):
+        return EclSum(str(pathlib.Path(runpath) / pathlib.Path(ecl_base)))
+    return None
```
wouterjdb, line 194: You are allowing the function to return a `None`. This means you will have to update the documentation and the type annotation to reflect this.

`def _load_simulations(runpath: str, ecl_base: str) -> Optional[EclSum]:`

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -243,7 +243,7 @@ def compute_metric_ensemble(obs, list_ensembles, metrics, str_key, iteration):
     return dict_metric
 
 
-def make_dataframe_simulation_data(path, eclbase_file, keys):
+def make_dataframe_simulation_data(path, eclbase_file, keys, end_date):
```
wouterjdb, line 246: While you're add it: write down the expected input types and return type.

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -271,35 +272,43 @@ def make_dataframe_simulation_data(path, eclbase_file, keys):
...
-            for counter, k in enumerate(keys):
-                if counter == 0:
-                    key_list_data = [
-                        x for x in realizations_dict[runpath] if x.startswith(k)
-                    ]
-                else:
-                    key_list_data.extend(
-                        [x for x in realizations_dict[runpath] if x.startswith(k)]
-                    )
+        if (realizations_dict[runpath] is not None) and (
```
wouterjdb, line 278:
```suggestion
        if realizations_dict[runpath] and (
```

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -271,35 +272,43 @@ def make_dataframe_simulation_data(path, eclbase_file, keys):
...
+            np.datetime64(realizations_dict[runpath].dates[-1]) >= end_date
+        ):
+            dates = realizations_dict[runpath].dates
+            if id_real == 0:
+                df_sim["DATE"] = pd.Series(dates)
+                df_sim["REAL_ID"] = pd.Series(id_real * np.ones(len(dates)), dtype=int)
+            df_tmp["DATE"] = pd.Series(dates)
+            df_tmp["REAL_ID"] = pd.Series(id_real * np.ones(len(dates)), dtype=int)
```
wouterjdb, line 286:
```suggestion
            df_sim["DATE"] = pd.Series(realizations_dict[runpath].dates)
            df_sim["REAL_ID"] = id_real 
```

if df_tmp is need you could do

`df_tmp = df_sim.copy(deep=True)`

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -271,35 +272,43 @@ def make_dataframe_simulation_data(path, eclbase_file, keys):
...
+                        key_list_data.extend(
+                            [x for x in realizations_dict[runpath] if x.startswith(k)]
+                        )
+
+            for key in key_list_data:
+                if id_real == 0:
+                    df_sim[key] = pd.Series(
+                        realizations_dict[runpath].numpy_vector(key)
+                    )
+                df_tmp[key] = pd.Series(realizations_dict[runpath].numpy_vector(key))
```
wouterjdb, line 304: Not sure why you are using the tmp df at all?

```suggestion
            for key in key_list_data:
                df_sim[key] = pd.Series(
                    realizations_dict[runpath].numpy_vector(key)
                )

```

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -271,35 +272,43 @@ def make_dataframe_simulation_data(path, eclbase_file, keys):
...
+
+            if id_real > 0:
+                df_sim = df_sim.append(df_tmp, ignore_index=True)
```
wouterjdb, line 307: If you don't need it after all then this could be removed.
```suggestion

```

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -271,35 +272,43 @@ def make_dataframe_simulation_data(path, eclbase_file, keys):
...
 
-        if id_real > 0:
-            df_sim = df_sim.append(df_tmp, ignore_index=True)
+            id_real = id_real + 1
 
-    return df_sim, realizations_dict, iteration
+    return df_sim, iteration, id_real
```
wouterjdb, line 311: Documentation isn't correct anymore now.

src/flownet/ahm/_ahm_iteration_analytics.py
```diff
@@ -381,10 +390,11 @@ def save_iteration_analytics():
     metrics = list(args.metrics.replace("[", "").replace("]", "").split(","))
 
     # Load ensemble of FlowNet
-    (df_sim, realizations_dict, iteration,) = make_dataframe_simulation_data(
+    (df_sim, iteration, nb_real) = make_dataframe_simulation_data(
         args.runpath,
         args.eclbase,
         list(args.quantity.replace("[", "").replace("]", "").split(",")),
+        np.datetime64(args.end),
```
wouterjdb, line 397: Why work with `np.datetime64` and not just with standard python dates?

wouterjdb (changes requested):
Some comments here and there 💬.

While you are updating this file anyways it could maybe be a good time to make sure you add type annotations on all functions.

wouterjdb:
While running a test I noticed also that the `upper` transformation might lead to problems when you don't use quotes in the vector names in the analysis config:

image

src/flownet/ahm/_ahm_iteration_analytics.py
edubarrosTNO: this type is used because that's the type for the dates when loading data from the summary file, and in order to filter the created datasets the dates type must be the same

src/flownet/ahm/_ahm_iteration_analytics.py
edubarrosTNO: I tried the suggested changes and the code no longer works. We need df_tmp because each realization might have different dates due to numerical solution of the simulations. So using the auxiliary df_tmp was the way that I found to be able to load all the data into the same dataframe

src/flownet/ahm/_ahm_iteration_analytics.py
edubarrosTNO: `df_tmp` is not just a copy, it's a temporary auxiliary dataframe to construct `df_sim`, which is a stacked dataframe

wouterjdb (approved):
LGTM 🚀

PR #268, opened 2020-11-26
Check $FLOW_PATH in environment variables and check `which flow`

tayloris:
…path

---

### Contributor checklist

- [x] :tada: This PR closes ##254
   - [x] Check if $FLOW_PATH is set, if so, use that. 
   - [x] If $FLOW_PATH points to something that doesn't exist, fail hard.
   - [x] Check $PATH for flow, 
       - [x] if $PATH for flow not found, fail hard with a message that flow was not found
       - [x] If FAIL include a message that either flow needs to be installed (link?) 
       - [x] or the user needs to set $FLOW_PATH manually

### DIFF
--- a/src/flownet/ert/_flow_job.py
+++ b/src/flownet/ert/_flow_job.py
@@ -29,15 +29,15 @@ def run_flow():
 
     flow_path = shutil.which("flow")
     if "FLOW_PATH" in os.environ:
-        flow_path = os.environ.get("FFFLOW_PATH")
+        flow_path = os.environ.get("FLOW_PATH")
         if not os.path.isfile(flow_path):
             raise AssertionError(
-                "FLOW_PATH does not point at a file that exists.\n \
+                r"FLOW_PATH does not point at a file that exists.\n \
                 Please, use the environment variable $FLOW_PATH to indicate a path for OPM\Flow"
             )
     elif flow_path is None:
         raise AssertionError(
-            "OPM/flow is not installed.\n \
+            r"OPM/flow is not installed.\n \
             Follow instructions in https://opm-project.org/ to install flow.\n \
             If OPM/flow it is already installed, make sure it is available in $PATH,\n \
             or alternatively use the environment variable $FLOW_PATH"
--- a/src/flownet/ert/_flow_job.py
+++ b/src/flownet/ert/_flow_job.py
@@ -2,6 +2,7 @@ import os
 import argparse
 import subprocess
 from pathlib import Path
+import shutil
 
 from ert_shared.plugins.plugin_response import plugin_response
 from ert_shared.plugins.plugin_manager import hook_implementation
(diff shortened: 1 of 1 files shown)


src/flownet/ert/_flow_job.py
```diff
@@ -26,8 +27,27 @@ def run_flow():
...
-    subprocess.run(
-        [os.environ.get("FLOW_PATH", "/usr/bin/flow"), args.data_file], check=True
-    )
+    if "FLOW_PATH" in os.environ:
+        flow_path = os.environ.get("FLOW_PATH")
+        if os.path.isfile(flow_path):
+            # Runing flow from ENV variable
+            subprocess.run([flow_path, args.data_file], check=True)
+        else:
+            raise AssertionError("FLOW_PATH points to a path that doesn't exist")
```
anders-kiaer, line 36:
```suggestion
            raise AssertionError("FLOW_PATH does not point at a file that exists")
```

src/flownet/ert/_flow_job.py
```diff
@@ -26,8 +27,27 @@ def run_flow():
...
+    elif shutil.which("flow") is None:
+        raise AssertionError(
+            "OPM/flow is not installed.\
+                             Follow instructions in https://opm-project.org/ to install flow."
+        )
```
anders-kiaer, line 41: Mention here that _if_ it is already installed, the user should make sure it is available in `$PATH` (or alternatively use the environment variable `$FLOW_PATH`).

src/flownet/ert/_flow_job.py
```diff
@@ -26,8 +27,27 @@ def run_flow():
...
+    else:
+        flow_path = shutil.which("flow")
+        if os.path.isfile(flow_path):
```
anders-kiaer, line 44: I think we don't need this check. If `shutil.which("flow") is not None` (which is the case here) it should already point to a file?

anders-kiaer (changes requested):
LGTM.

Maybe restructure the code slightly? Currently `subprocess.run(...)` is one multiple lines.

Could instead 
1) First find a valid path to `flow` executable (and store it in a variable), and do any error raising.
2) Run subprocess with the found `flow` executable.

This will also move some code out of the `if/else` blocks.

src/flownet/ert/_flow_job.py
tayloris: Yes, thats true

tayloris:
I included comments from  @anders-kiaer . The entry point to ERT is defined in `setup.py` so I think the best place to inform the user about  $FLOW_PATH is in the `README` for flownet installation. The implementation will raise an error informing about $FLOW_PATH if there is any problem when we want to run flow. 

If we want to raise a warning about $FLOW_PATH  we could do it in main() but I'm not sure about it. Let me know what you think.

src/flownet/ert/_flow_job.py
```diff
@@ -26,8 +27,22 @@ def run_flow():
...
+            )
+    elif flow_path is None:
+        raise AssertionError(
+            r"OPM/flow is not installed.\n \
+            Follow instructions in https://opm-project.org/ to install flow.\n \
+            If OPM/flow it is already installed, make sure it is available in $PATH,\n \
+            or alternatively use the environment variable $FLOW_PATH"
+        )
+
+    subprocess.run([flow_path, args.data_file], check=True)
```
wouterjdb, line 46:
```suggestion
    flow_path = shutil.which("flow")
    if "FLOW_PATH" in os.environ:
        flow_path = os.environ.get("FLOW_PATH")
        if not os.path.isfile(flow_path):
            raise FileNotFoundError(
                r"$FLOW_PATH does not point at a file that exists.\n \
                Please, use the environment variable $FLOW_PATH to indicate a path for OPM Flow"
            )
    elif flow_path is None:
        raise RuntimeError(
            r"OPM Flow could not be found.\n \
            Follow instructions on https://opm-project.org/ to install OPM Flow.\n \
            If OPM Flow it is already installed, make sure it is available in $PATH,\n \
            or alternatively use the environment variable $FLOW_PATH."
        )

    subprocess.run([flow_path, args.data_file], check=True)
```

wouterjdb (changes requested):
LGTM, minor suggestions.

wouterjdb (approved):
LGTM 👍

anders-kiaer (approved):

PR #272, opened 2020-11-30
Adds resampling of observations

edubarrosTNO:
*Adds resampling of observation dates at requested frequency by finding the nearest date among existing observation dates.*

---

### Contributor checklist

- [x] :tada: This PR partially addresses #206.
   - [x] Resampling observation dates based on requested frequency.
   - [x] Adds (optional) resampling entry to the config parser.


src/flownet/ert/_create_ert_setup.py
```diff
@@ -47,14 +48,42 @@ def create_observation_file(
         Nothing
 
     """
-    num_dates = len(schedule.get_dates())
+    # pylint: disable-msg=too-many-locals
+    dt_schedule = pd.to_datetime(schedule.get_dates())
```
wouterjdb, line 52: Consider renaming a dataframe to `df_xxxx` (not `dt_`)

src/flownet/ert/_create_ert_setup.py
```diff
@@ -47,14 +48,42 @@ def create_observation_file(
...
+    # no interpolation: keeps nearest existing date
+    if config.flownet.data_source.simulation.resampling:
+        dt_resampled = pd.date_range(
+            schedule.get_dates()[1],
+            schedule.get_dates()[-1],
+            freq=config.flownet.data_source.simulation.resampling,
+        )
+        df_schedule = pd.DataFrame(data=range(len(dt_schedule)), index=dt_schedule)
+        idx = np.zeros(len(dt_resampled), dtype=int)
+        for i, k in enumerate(dt_resampled):
```
wouterjdb, line 64: How fast is this loop?

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -19,6 +19,8 @@
 _MIN_ERROR = 10
 _REL_ERROR = 0.05
 
+_RESAMPLING = None
```
wouterjdb, line 22: Instead of never testing the resampling, it would be better to create a test case that tests resampling based on the data provided by @tayloris when he created the original test.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -245,14 +251,41 @@ def test_check_obsfiles_ert_yaml() -> None:
...
+        )
+        df_schedule = pd.DataFrame(data=range(len(dt_schedule)), index=dt_schedule)
+        idx = np.zeros(len(dt_resampled), dtype=int)
+        for i, k in enumerate(dt_resampled):
+            idx[i] = df_schedule.index.get_loc(k, method="nearest")
+        dates = [
+            d.date() for d in df_schedule.iloc[np.unique(idx)].index.to_pydatetime()
+        ]
+    else:
+        dates = schedule.get_dates()
```
wouterjdb, line 272: The test should then actually test this code - do not re-write the exact same code here (rather make it into a function in the other file and test whether the function returns what you expect).

src/flownet/config_parser/_config_parser.py
```diff
@@ -201,6 +201,12 @@ def _to_abs_path(path: Optional[str]) -> str:
                                             },
                                         },
                                     },
+                                    "resampling": {
+                                        MK.Type: types.String,
+                                        MK.AllowNone: True,
+                                        MK.Description: "Requested observation frequency, used e.g. weekly (W), "
```
wouterjdb, line 207: Maybe mention that all standard pandas resampling is supported?

src/flownet/ert/_create_ert_setup.py
```diff
@@ -47,14 +48,42 @@ def create_observation_file(
...
-        ["_training", 1, num_training_dates],
-        ["_test", num_training_dates + 1, num_dates],
-    ]
+    if config.flownet.data_source.simulation.resampling:
+        export_settings = [
+            ["_complete", 0, num_dates],
+            ["_training", 0, num_training_dates],
+            ["_test", num_training_dates + 1, num_dates],
+        ]
+    else:
```
wouterjdb, line 81: This could potentially be removed if counting from 0 is possible.

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -245,14 +251,41 @@ def test_check_obsfiles_ert_yaml() -> None:
...
-        ["_training", 1, num_training_dates],
-        ["_test", num_training_dates + 1, num_dates],
-    ]
+    if config.flownet.data_source.simulation.resampling:
+        export_settings = [
+            ["_complete", 0, num_dates],
+            ["_training", 0, num_training_dates],
+            ["_test", num_training_dates + 1, num_dates],
+        ]
+    else:
```
wouterjdb, line 283: This could potentially be removed if counting from zero is possible.

wouterjdb (changes requested):

src/flownet/ert/_create_ert_setup.py
edubarrosTNO: it is not slow as it only concerns the dates, not all the data. I tried to avoid the for-loop, but the `get_loc` method of `pandas` isn't vectorized

src/flownet/ert/_create_ert_setup.py
```diff
@@ -47,14 +48,42 @@ def create_observation_file(
...
     """
-    num_dates = len(schedule.get_dates())
+    # pylint: disable-msg=too-many-locals
+    dt_schedule = pd.to_datetime(schedule.get_dates())
+
+    # Resampling dates based on requested frequency with
+    # no interpolation: keeps nearest existing date
+    if config.flownet.data_source.simulation.resampling:
+        dt_resampled = pd.date_range(
+            schedule.get_dates()[1],
```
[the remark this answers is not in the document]
edubarrosTNO: As discussed this morning, there is a reason for skipping the first observation date, it has to do with which dates OPM-Flow simulations export (first date included or not). @olelod should have more information about it

tests/test_check_obsfiles_ert_yaml.py
```diff
@@ -100,6 +102,122 @@ def compare(ert_obs_dict: dict, yaml_obs_dict: dict) -> None:
             assert yaml_obs[list_item["key"]][2] == ert_obs_dict[list_item["key"]][2]
 
 
+def create_schedule_from_data(
```
wouterjdb, line 105: This is an internal function, I suggest to use a leading underscore to empathize that. 

```suggestion
def _create_schedule_from_data(
```

wouterjdb (changes requested):
LGTM. Two minor comments:
1) See review, I suggest to indicate that this new function is an internal function by using an underscore. This also makes it clearer for the reader it is not really a test as such.
2) You now always test with resampling. Could you maybe run, in that same test, twice the same function. Once with resampling (and check that the output is as expect) and once without resampling (and check that the output is as expected).

src/flownet/ert/_create_ert_setup.py
```diff
@@ -23,6 +24,42 @@
...
+        )
+        df_schedule = pd.DataFrame(
+            data=range(len(df_dates_schedule)), index=df_dates_schedule
+        )
+        idx = np.zeros(len(df_dates_resampled), dtype=int)
+        for i, k in enumerate(df_dates_resampled):
+            idx[i] = df_schedule.index.get_loc(k, method="nearest")
+        dates = [
+            d.date() for d in df_schedule.iloc[np.unique(idx)].index.to_pydatetime()
+        ]
```
wouterjdb, line 56: Should you check if the first resampled date is equal to the original first date; if so: remove it!?

edubarrosTNO: this is not needed, because it will only loop through the range of dates starting at the second date

wouterjdb (approved):
LGTM 🚀

PR #283, opened 2020-12-04
Change absolute path to relative path for some variables in ERT

### LINKED ISSUE #274: Full path used in some variables in ERT config
The following paths are written as absolute paths:

https://github.com/equinor/flownet/blob/d00960b44c561e2ce5e5630f278179b25ac59241/src/flownet/templates/common_config.ert.jinja2#L22-L24

This should be changed to relative paths.

tayloris:
When we render each Eclipse/OPM simulation file for each realization we require several file name arguments that are currently absolute paths. 

Each realization is to be render inside `OUTPUT_FOLDER/output/runpath/realization-*/iter-*`, and the files we need to pass as arguments live in OUTPUT_FOLDER/.

I removed the absolute path and include "../../../../" to point towards the corresponding parent directory.

---

### Contributor checklist

- [x] :tada: This PR closes #274.
   - [x] Change absolute paths to relative path

### DIFF
--- a/src/flownet/ert/_create_ert_setup.py
+++ b/src/flownet/ert/_create_ert_setup.py
@@ -176,9 +176,9 @@ def create_ert_setup(  # pylint: disable=too-many-arguments
         analysis_quantity = str(None)
 
     configuration = {
-        "pickled_network": output_folder.resolve() / "network.pickled",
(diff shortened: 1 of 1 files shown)


wouterjdb:
#274 is no longer an open issue.

PR #284, opened 2020-12-07
add totals as measurements

### LINKED ISSUE #266: Use cumulative production as observations
Currently we see that the cumulative production is always off. Could it help steering the model in the right direction by condition the model to cumulative "observations" as well?

olwijn:

---

### Contributor checklist

- [x] :tada: This PR closes #266.
   - [x] Provide support for cumulative rates as observation data
   - [x] Tested the use of cumulate rate data at only the final time


src/flownet/realization/_schedule.py
```diff
@@ -187,6 +187,9 @@ def _calculate_wconhist(self):
                         oil_rate=value["WOPR"],
                         water_rate=value["WWPR"],
                         gas_rate=value["WGPR"],
+                        oil_total=value["WOPT"],
```
wouterjdb, line 190: Recalculating could potentially mean that one doesn't have to supply this at all.

src/flownet/data/from_flow.py
```diff
@@ -153,7 +156,7 @@ def _production_data(self) -> pd.DataFrame:
             * Improve robustness pf setting of Phase and Type.
 
         """
-        keys = ["WOPR", "WGPR", "WWPR", "WBHP", "WTHP", "WGIR", "WWIR", "WSTAT"]
+        keys = ["WOPR", "WGPR", "WWPR", "WOPT", "WGPT", "WWPT", "WBHP", "WTHP", "WGIR", "WWIR", "WGIT", "WWIT", "WSTAT"]
```
wouterjdb, line 159: You're adding `WGIT` and `WWIT` here but not in the documentation a few lines earlier and neither are they added in `src/flownet/config_parser/_config_parser.py`. Is that intentional?

Should you read those values at all? You can reconstruct the cumulative at any given date by calculating the cumulative sum over the rates (this would also be easier to use later on in other data sources)? Or do we expect big differences based on differences in reporting steps?

olelod: The rates are instantaneous, and depending on the report step length we can get a quite different result reading the actual cumulative volumes and calculating them from rates. Over the years when doing uncertainty studies I have always calculated average yearly or monthly rates from cumulative volumes - for history matching purposes I guess the instantaneous rates are important - what I am saying is that we may need to import both then.

olwijn: We may indeed not be using WGIT and WWIT, unless we prescribe BHP instaed of the rates during the history matching. I thought it would be good for completeness to just add them. I forgot about adding them to **src/flownet/config_parser/_config_parser.py.**

src/flownet/config_parser/_config_parser.py
```diff
@@ -136,6 +136,51 @@ def _to_abs_path(path: Optional[str]) -> str:
                                                     },
                                                 },
                                             },
+                                            "WOPT": {
```
olelod, line 139: When adding WOPT (or WGPT or WWPT), should we allow for also adding WOPR (or WGPR or WWPR)? Conditioning on both rates and cumulatives i guess is kind of a messy thing with all the correlation going on. Should we add a check in the config_parser to make sure both are not defined?

olelod (changes requested):
I guess we should also add to the CHANGELOG that it is possible to condition on cumulatives now. Other than that it looks good to me!

src/flownet/config_parser/_config_parser.py
olwijn: I will make a new issue suggesting this and add an item to the CHANGELOG

olelod (approved):
LGTM

PR #298, opened 2021-01-07
Improved model generation

### LINKED ISSUE #282: Long connections across Norne separate regions to be removed
While inspecting the grid of one of the FlowNet realizations of Norne, it was observed that extremely long connections are appearing between well nodes in the two "tentacles" of the model. A fixed maximum distance value should be used to determine connections to be removed instead of a fraction as it is currently set up

image

### LINKED ISSUE #297: Expose angle removal constraint for connections to user
We currently have a hard-coded constraint that a triangle create by the Delaunay triangulation will removed if it has an angle larger than 150 degrees.

https://github.com/equinor/flownet/blob/6eb0fd57d1a328567a46b68b0afdf17aa6117f56/src/flownet/network_model/_generate_connections.py#L50

This seems to lead to a lot of duplicate connections in the case of multiple connections per well:

image

After changing the angle to 80 (maybe a bit too strict), the same area looked like:

image

By letting the user tweak this number the number of duplicate connections could be reduced significantly.

wouterjdb:
This PR fixes the issues described in #282 and #297.

---

### Contributor checklist

- [x] :tada: This PR closes #282 and closes #297.
   - [x] Fix long connection removal in non-reservoir
   - [x] Expose angle to user via config file
   - [x] Expose non-reservoir removal parameter _n_ to user
   - [ ] ~~Find good hyper parameters for test-data set~~


wouterjdb:
Currently this PR has implemented a smaller or larger than check for the angle - I'm not so sure this is actually the right way @olelod. I think what it would do is the following:

With only the larger than x you would remove:
image

If you also make it into a smaller than test, you'd remove:
image

That is not really what we are after?

wouterjdb:
Conclusion discussion: @olelod @LonnekevB 

Only expose angle, don't do the rest yet. Make new issue.

src/flownet/network_model/_generate_connections.py
```diff
@@ -309,6 +319,8 @@ def _create_entity_connection_matrix(
     aquifer_ends: List[Coordinate],
     max_distance_fraction: float,
     max_distance: float,
+    concave_hull_bounding_boxes: Optional[np.ndarray] = None,
+    n_non_reservoir_evaluation: Optional[int] = 10,
```
olelod, line 323: When the number of points is set as default in the config file, will it ever not be defined?

olelod (approved):
LGTM

src/flownet/network_model/_generate_connections.py
wouterjdb: Unless you call the function via some other way.

PR #304, opened 2021-01-15
Enable new layer functionality in FlowNet

wouterjdb:
This ties together all recent work related to layering.

---

### Contributor checklist

- [x] :tada: This PR closes #276 (and will close #239).
- [x] :scroll: I have broken down my PR into the following tasks:
   - [x] Implement final layering functionality
   - [ ] Fix type errors for concave hull code
   - [ ] Decide how to split additional nodes
- [ ] :robot: I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR.
- [ ] :book: I have considered adding a new entry in `CHANGELOG.md`.
- [ ] :books: I have considered updating the documentation.

PR #305, opened 2021-01-20
I276 enabled new layering functionality

### LINKED ISSUE #276: Allow for multiple separated network in a single FlowNet
This issue is to support a stepwise implementation of a layered FlowNet approach. This issue describes a feature that allows for the generation of 2 or more independent network in a single FlowNet.

**What**
Allow for multiple networks to be generated in a single FlowNet that are only communicating via well connections. This can be achieved relatively straight forward with the current code and can serve as a proof of concept that history matching performance could be improved by a layered approach.

**How**

-----

The first three tasks are addressed in issue #277, PR #279.

- [x] Allow to specify layering in the FlowNet configuration file, for example:
```yaml
flownet:
    data_source:
        simulation:
            layering:
               - [ 1, 3 ]
               - [ 4, ... ]
```

- [x] Translate the layering for each well into depth intervals. The depth intervals are what will need to be used to assign layer_id's to each well connection (this to propagate simulation specific information in the code).

- [x] Create a function that assign layer_id's to well connections. For example, this could be added to `field_data.well_connections` as an additional column.
----

This task is addressed in PR #287.

- [x] A function needs to be created that allows for creating the `_grid_cell_bounding_box` for a specific layer. This can be done by extending this function:
https://github.com/equinor/flownet/blob/d00960b44c561e2ce5e5630f278179b25ac59241/src/flownet/data/from_flow.py#L283

----

- [x] Update perforation strategy PR #296 

----

PR #304.

- [ ] In __generate_connections.py_ the `create_connections()` could be expanded to allow for splitting models based on depth intervals per well. That is, an additional input "layer_id" could be supplied that identifies which well coordinates in `df_coordinates` belong to which zones.

- [ ] By using this information, as well as updating the concave_hull_bounding_box, one could now run the `_generate_connections` function multiple times (which in turn performs Mitchells Best Candidate and Delaunay triangulation). This will generate multiple lists of start and end points for all connections in the FlowNet. Some rules for dividing the additional nodes over the various network need to be used (for example divide based on number of connections associated with a layer_id).

LonnekevB:

---

### Contributor checklist

- [ ] :tada: This PR closes #276 (and closes #304).
   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/src/flownet/ahm/_run_ahm.py
+++ b/src/flownet/ahm/_run_ahm.py
@@ -419,13 +419,10 @@ def run_flownet_history_matching(
     concave_hull_list: Optional[List[np.ndarray]] = None
     if config.flownet.data_source.concave_hull:
         concave_hull_list = []
-        if df_well_connections["LAYER_ID"].nunique() > 1:
-            for layer_id in df_well_connections["LAYER_ID"].unique():
-                concave_hull_list.append(
-                    field_data.grid_cell_bounding_boxes(layer_id=layer_id)
-                )
-        else:
-            concave_hull_list.append(field_data.grid_cell_bounding_boxes())
+        for layer_id in df_well_connections["LAYER_ID"].unique():
+            concave_hull_list.append(
+                field_data.grid_cell_bounding_boxes(layer_id=layer_id)
+            )
 
     df_entity_connections: pd.DataFrame = create_connections(
         df_well_connections[["WELL_NAME", "X", "Y", "Z", "LAYER_ID"]].drop_duplicates(
--- a/src/flownet/ahm/_run_ahm.py
+++ b/src/flownet/ahm/_run_ahm.py
@@ -417,12 +417,21 @@ def run_flownet_history_matching(
         field_data.faults if config.model_parameters.fault_mult else None
     )
 
-    concave_hull_bounding_boxes: Optional[np.ndarray] = None
+    concave_hull_bounding_boxes: Optional[List[np.ndarray]] = None
     if config.flownet.data_source.concave_hull:
-        concave_hull_bounding_boxes = field_data.grid_cell_bounding_boxes
+        concave_hull_bounding_boxes = []
+        if df_well_connections["LAYER_ID"].nunique() > 1:
+            for layer_id in df_well_connections["LAYER_ID"].unique():
+                concave_hull_bounding_boxes.append(
+                    field_data.grid_cell_bounding_boxes(layer_id=layer_id)
+                )
+        else:
+            concave_hull_bounding_boxes.append(field_data.grid_cell_bounding_boxes())
 
     df_entity_connections: pd.DataFrame = create_connections(
-        df_well_connections[["WELL_NAME", "X", "Y", "Z"]].drop_duplicates(keep="first"),
+        df_well_connections[["WELL_NAME", "X", "Y", "Z", "LAYER_ID"]].drop_duplicates(
+            keep="first"
+        ),
         config,
         concave_hull_bounding_boxes=concave_hull_bounding_boxes,
     )
--- a/src/flownet/ahm/_run_ahm.py
+++ b/src/flownet/ahm/_run_ahm.py
@@ -417,23 +417,23 @@ def run_flownet_history_matching(
         field_data.faults if config.model_parameters.fault_mult else None
     )
 
-    concave_hull_bounding_boxes: Optional[List[np.ndarray]] = None
+    concave_hull_list: Optional[List[np.ndarray]] = None
     if config.flownet.data_source.concave_hull:
-        concave_hull_bounding_boxes = []
+        concave_hull_list = []
         if df_well_connections["LAYER_ID"].nunique() > 1:
             for layer_id in df_well_connections["LAYER_ID"].unique():
-                concave_hull_bounding_boxes.append(
+                concave_hull_list.append(
                     field_data.grid_cell_bounding_boxes(layer_id=layer_id)
                 )
         else:
-            concave_hull_bounding_boxes.append(field_data.grid_cell_bounding_boxes())
+            concave_hull_list.append(field_data.grid_cell_bounding_boxes())
 
     df_entity_connections: pd.DataFrame = create_connections(
         df_well_connections[["WELL_NAME", "X", "Y", "Z", "LAYER_ID"]].drop_duplicates(
             keep="first"
         ),
         config,
-        concave_hull_bounding_boxes=concave_hull_bounding_boxes,
+        concave_hull_list=concave_hull_list,
     )
 
     network = NetworkModel(
--- a/src/flownet/config_parser/_config_parser.py
+++ b/src/flownet/config_parser/_config_parser.py
@@ -1614,6 +1614,12 @@ def parse_config(
                 f"Valid way to define the layering is as nested list with "
                 f"layer intervals. e.g. ((1, 5),(6, 9),(10, 15))."
             )
+    if layers and not config.flownet.data_source.concave_hull:
+        raise ValueError(
+            f"Concave_hull should be True when layers are defined. "
+            f"The concave hulls of the layers are used to split"
+            f"the number of additional nodes between the layers."
+        )
 
     req_relp_parameters: List[str] = []
     if (
--- a/src/flownet/config_parser/_config_parser.py
+++ b/src/flownet/config_parser/_config_parser.py
@@ -1624,7 +1624,7 @@ def parse_config(
     if layers and not config.flownet.data_source.concave_hull:
         raise ValueError(
             "Concave_hull should be True when layers are defined. "
-            "The concave hulls of the layers are used to split"
+            "The concave hulls of the layers are used to split "
             "the number of additional nodes between the layers."
         )
 
--- a/src/flownet/config_parser/_config_parser.py
+++ b/src/flownet/config_parser/_config_parser.py
@@ -1616,9 +1616,9 @@ def parse_config(
             )
     if layers and not config.flownet.data_source.concave_hull:
         raise ValueError(
-            f"Concave_hull should be True when layers are defined. "
-            f"The concave hulls of the layers are used to split"
-            f"the number of additional nodes between the layers."
+            "Concave_hull should be True when layers are defined. "
+            "The concave hulls of the layers are used to split"
+            "the number of additional nodes between the layers."
         )
 
     req_relp_parameters: List[str] = []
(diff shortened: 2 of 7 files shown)


src/flownet/network_model/_generate_connections.py
```diff
@@ -94,9 +94,46 @@ def _remove_recorded_connections(
     return conn_matrix
 
 
+def _split_additional_flow_nodes(
+    configuration: Any, concave_hull_bounding_boxes: np.ndarray
```
wouterjdb, line 98: I suggest to limit the input to the additional flow nodes; not the entire configuration. This makes the function more "stand alone" and easier to test.

src/flownet/network_model/_generate_connections.py
```diff
@@ -94,9 +94,46 @@ def _remove_recorded_connections(
...
+    fractions = [v / sum(volumes) for v in volumes]
+    sep_add_flownodes = [round(frac * total_additional_nodes) for frac in fractions]
+
+    return sep_add_flownodes
+
+
 def _generate_connections(
     df_coordinates: pd.DataFrame,
     configuration: Any,
+    additional_flownodes: int,
```
wouterjdb, line 136: Here you add the `additional_flownodes`, because you have already split them before going into this function.

However, as you also pass the configuration and the `concave_hull_bounding_boxes list` you already have all required info and you could do the splitting of additional nodes _inside_ this function, isn't it?

Oh by the way, maybe call this variable `additional_flow_nodes` to be aligned with underscores in earlier naming...

src/flownet/network_model/_generate_connections.py
```diff
@@ -319,7 +357,7 @@ def _create_entity_connection_matrix(
     aquifer_ends: List[Coordinate],
     max_distance_fraction: float,
     max_distance: float,
-    concave_hull_bounding_boxes: Optional[np.ndarray] = None,
+    concave_hull_bounding_boxes: Optional[List[np.ndarray]] = None,
```
wouterjdb, line 360: I changed `concave_hull_bounding_boxes` into a list of _concave_hull_bounding_boxes_. So the naming is not good anymore. Maybe it should be called something like _concave_hull_set_?

src/flownet/network_model/_generate_connections.py
LonnekevB: Good point, will do!

src/flownet/network_model/_generate_connections.py
wouterjdb: eh, maybe not "set" if it is a "list" by the way.... then we should use a set ;) You choose 👍

LonnekevB:
image
 The CI/CD now crashes on this. Don't really understand why it is even referring to a lambda function. Any ideas @wouterjdb ?

src/flownet/network_model/_generate_connections.py
LonnekevB: concave_hull_list it is ;)

wouterjdb:
> The CI/CD now crashes on this. Don't really understand why it is even referring to a lambda function. Any ideas @wouterjdb ?

You are creating a collection called `additional_flow_nodes` and setting it to a value `=2` one line later?

LonnekevB:
> You are creating a collection called `additional_flow_nodes` and setting it to a value `=2` one line later?

But that was already there. The only thing I added is adding additional_flownodes to the _generate_connections(): 

`    starts, ends = _generate_connections(
        df_coordinates=DF_COORDINATES,
        configuration=config,
        additional_flownodes=config.flownet.additional_flow_nodes,
    )
`

wouterjdb:
If the test actually runs fine you could consider ignoring the pylint warning: `# pylint: disable=no-member.`

src/flownet/ahm/_run_ahm.py
```diff
@@ -416,14 +416,23 @@ def run_flownet_history_matching(
         field_data.faults if config.model_parameters.fault_mult else None
     )
 
-    concave_hull_bounding_boxes: Optional[np.ndarray] = None
+    concave_hull_list: Optional[List[np.ndarray]] = None
     if config.flownet.data_source.concave_hull:
-        concave_hull_bounding_boxes = field_data.grid_cell_bounding_boxes
+        concave_hull_list = []
+        if df_well_connections["LAYER_ID"].nunique() > 1:
```
wouterjdb, line 422: Is the if-statement actually needed? (Are LayerID's also assigned when there is only 1 layer; if so you could drop the if statement and always go through the loop)

src/flownet/config_parser/_config_parser.py
```diff
@@ -1621,6 +1621,12 @@ def parse_config(
                 f"Valid way to define the layering is as nested list with "
                 f"layer intervals. e.g. ((1, 5),(6, 9),(10, 15))."
             )
+    if layers and not config.flownet.data_source.concave_hull:
+        raise ValueError(
+            "Concave_hull should be True when layers are defined. "
+            "The concave hulls of the layers are used to split"
```
wouterjdb, line 1627:
```suggestion
            "The concave hulls of the layers are used to split "
```

src/flownet/network_model/_generate_connections.py
```diff
@@ -94,9 +94,51 @@ def _remove_recorded_connections(
     return conn_matrix
 
 
+def _split_additional_flow_nodes(
+    total_additional_nodes: int, concave_hull_list: List[np.ndarray]
+) -> List:
```
wouterjdb, line 99:
```suggestion
) -> List[int]:
```

src/flownet/network_model/_generate_connections.py
```diff
@@ -94,9 +94,51 @@ def _remove_recorded_connections(
...
+    """
+    This function takes splits the additional_flow_nodes defined in the config over the layers.
```
wouterjdb, line 101:
```suggestion
    This function splits the additional_flow_nodes defined in the config over the layers.
```

src/flownet/network_model/_generate_connections.py
```diff
@@ -94,9 +94,51 @@ def _remove_recorded_connections(
...
+    The division is based on the size (volume) of the boundingbox of the layer.
```
wouterjdb, line 102:
```suggestion
    The division is based on the sum of the volume of the boundingboxes in the layer.
```

src/flownet/network_model/_generate_connections.py
```diff
@@ -94,9 +94,51 @@ def _remove_recorded_connections(
...
+
+    Args:
+        configuration: FlowNet configuration yaml as dictionary
```
wouterjdb, line 105:
```suggestion
        total_additional_nodes: The total number of additional nodes to add to the model
```

src/flownet/network_model/_generate_connections.py
```diff
@@ -94,9 +94,51 @@ def _remove_recorded_connections(
...
+        concave_hull_list: List of boundingbox per layer, i.e., numpy array with x, y, z min/max
```
wouterjdb, line 106:
```suggestion
        concave_hull_list: List of boundingboxes per layer, i.e., numpy array with x, y, z min/max
```

src/flownet/network_model/_generate_connections.py
```diff
@@ -94,9 +94,51 @@ def _remove_recorded_connections(
...
+            boundingboxes for each grid block
+
+    Returns:
+        List of additional notes per layer (len of list same as number of layers)
```
wouterjdb, line 110:
```suggestion
        List of additional nodes per layer (len of list, same as number of layers)
```

src/flownet/network_model/_generate_connections.py
```diff
@@ -94,9 +94,51 @@ def _remove_recorded_connections(
...
+    fractions = [v / sum(volumes) for v in volumes]
+    sep_add_flownodes = [round(frac * total_additional_nodes) for frac in fractions]
+
+    return sep_add_flownodes
+
+
 def _generate_connections(
     df_coordinates: pd.DataFrame,
     configuration: Any,
+    additional_flow_nodes: int,
```
wouterjdb, line 141: Why the change in name: additional **_flow** _nodes?

src/flownet/network_model/_generate_connections.py
```diff
@@ -106,7 +148,8 @@ def _generate_connections(
 
     Args:
         df_coordinates: coordinates on original DataFrame format
-        configuration: FlowNet configuration yaml
+        configuration: Flownet configuration yaml,
+        additional_flownodes: Number of additional flow nodes to generate
```
wouterjdb, line 152: Yet another name 😄 ❓

src/flownet/network_model/_generate_connections.py
```diff
@@ -333,7 +376,8 @@ def _create_entity_connection_matrix(
         aquifer_ends: List of coordinates of all aquifer ends
         max_distance_fraction: Fraction of longest connection distance to be removed
         max_distance: Maximum distance between nodes, removed otherwise
-        concave_hull_bounding_boxes: Numpy array with x, y, z min/max boundingboxes for each grid block
+        concave_hull_list: List of boundingbox per layer, i.e., numpy array with x, y, z min/max
```
wouterjdb, line 379:
```suggestion
        concave_hull_list: List of boundingboxes per layer, i.e., numpy array with x, y, z min/max
```

tests/test_generate_connections.py
```diff
@@ -104,9 +105,12 @@ def test_generate_connections() -> None:
...
     config.flownet.angle_threshold = None
 
+    # pylint: disable=no-member
     starts, ends = _generate_connections(
-        df_coordinates=DF_COORDINATES, configuration=config
-    )
+        df_coordinates=DF_COORDINATES,
+        configuration=config,
+        additional_flow_nodes=config.flownet.additional_flow_nodes,
+    )  # pylint: disable=no-member
```
wouterjdb, line 113: Remove pylint disable once?

wouterjdb (changes requested):

wouterjdb:
Also a an item in the CHANGELOG. 👍

src/flownet/ahm/_run_ahm.py
LonnekevB: I think you are right. We always assign a layerID, if no layering is defined they just all have the same layerID

src/flownet/network_model/_generate_connections.py
LonnekevB: Forgot to change this one indeed ;)

src/flownet/network_model/_generate_connections.py
LonnekevB: Because you suggested that earlier on ;) 
https://github.com/equinor/flownet/pull/305#discussion_r561673669

src/flownet/ahm/_run_ahm.py
wouterjdb: Ok, then its not needed (anymore).

wouterjdb (approved):

PR #322, opened 2021-02-03
Support for separate RSVD tables in equilibration regions

### LINKED ISSUE #318: Support to RSVD tables for multiple equilibration regions
In the Norne example we have 5 different equilibration regions and 3 different tables for dissolved gas (RSVD Eclipse keyword) in the system. Currently, the generated FlowNet models assign a single RSVD table to all 5 equilibration regions - this single RSVD table is prescribed in the `flownet-testdata/norne/config/norne_static/rsvd.csv` file. An enhancement would be to be able to prescribe different tables for different equilibration regions.

olelod:
Adding possibility for defining different RSVD tables for each EQLNUM region.

---

### Contributor checklist

- [x] :tada: This PR closes #318 .
   - [x] Add rsvd_multiple.csv to flownet-testdata (3 columns, depth, rs, eqlnum)
   - [x] Update render_output() to allow for 2 or 3 columns in csv input
   - [x] Update RSVD jinja template


.github/workflows/flownet.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           TESTDATA_REPO_OWNER: equinor
           # If you want the CI to (temporarily) run against another branch than master,
           # change the value her from "master" to the relevant branch name.
-          TESTDATA_REPO_BRANCH: master
+          TESTDATA_REPO_BRANCH: rsvd-tables
```
wouterjdb, line 46: Don't forget to revert.

src/flownet/parameters/_equilibration.py
```diff
@@ -125,10 +125,25 @@ def render_output(self) -> Dict:
             )
 
             if self._rsvd is not None:
+                df_rsvd = pd.read_csv(self._rsvd)
+                if len(df_rsvd.columns) == 3:
+                    df_rsvd.columns = ["depth", "rs", "eqlnum"]
```
wouterjdb, line 130: Here we assume that the columns are in this order. Would it be safer to let the user _name_ the columns. We then read the column names (case insensitive) and use those. If the column names are spelled wrongly we return an error saying: "Didn't find this or that".

That's maybe somewhat less error prone?

wouterjdb (changes requested):
Some minor comments...

src/flownet/templates/RSVD.jinja2
```diff
@@ -1,7 +1,12 @@
 RSVD
```
edubarrosTNO, line 1: please add blank line before RSVD keyword to improve readability of generated INCLUDE file

src/flownet/templates/RSVD.jinja2
```diff
@@ -1,7 +1,12 @@
...
+{% for i in range(1, nr_eqlnum + 1) -%}
+{%- if nr_rsvd > 1 -%}
+{%- for _, row in rsvd[rsvd['eqlnum']==i].iterrows() -%}
+{{ row['depth'] }} {{ row['rs'] }}
+{% endfor %}/
+{%- else -%}
 {%- for _, row in rsvd.iterrows() -%}
 {{ row['depth'] }} {{ row['rs'] }}
 {% endfor %}/
+{% endif %}
```
edubarrosTNO, line 11: there is one blank line appearing between the tables of the multiple regions when using a single RSVD table is provided in the `rsvd.csv` file. It due to one of the lines above in the jinja file

edubarrosTNO (changes requested):
Besides my minor comments and the suggestions from Wouter, LGTM !
I tested it locally and the code seems to do what it should.

src/flownet/templates/RSVD.jinja2
olelod: Added

src/flownet/templates/RSVD.jinja2
olelod: Re-wrote the code a bit so that there is no if-statement in the Jinja template file. Should be better now.

src/flownet/parameters/_equilibration.py
olelod: Added checking of the csv file in _config_parser.py, so when we arrive at this part of the code, the file should be known to be ok

.github/workflows/flownet.yml
olelod: Done

src/flownet/templates/RSVD.jinja2
```diff
@@ -1,7 +1,7 @@
+
```
wouterjdb, line 1: Why a leading enter?

wouterjdb (approved):

PR #341, opened 2021-02-26
Add option to remove flow path if permeability is below a threshold

### LINKED ISSUE #329: Remove low permeability tubes
Possibility to remove a tubes when its permeability value is very low. This should help solving the problem of simulation convergence problems.

wouterjdb:
This PR adds the option to remove flow path if permeability is below a threshold

---

### Contributor checklist

- [x] :tada: This PR closes #329.
   - [x] Add removal filter
   - [x] Expose option to user


wouterjdb:
Setting a threshold of 50 mD will result in (the grey tube that have cells where below 50 mD and are now deactivated):
image

edubarrosTNO:
> Setting a threshold of 50 mD will result in (the grey tubes that have cells where below 50 mD and are now deactivated):

Ok, so it looks like this threshold is indeed taking effect.
One question that I have is: how would you choose such threshold? I mean, we already define the minimum permeability value in the config file. Would this new threshold be equal to that minimum value or slightly higher? Where do we draw the line?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## Unreleased
 
 ### Added
+- [#341](https://github.com/equinor/flownet/pull/341) Added option to add a min permeability threshold for which tubes are deactivated when the permeability is below this value.
```
edubarrosTNO, line 8: Add a comment here on how this new threshold value relates to / should be used with the minimum permeability value defined in the config file. See my comment in the PR page

src/flownet/config_parser/_config_parser.py
```diff
@@ -437,6 +437,11 @@ def _to_abs_path(path: Optional[str]) -> str:
                         MK.Description: "Number of points along a tube to check whether they are in"
                         "non reservoir for removal purposes.",
                     },
+                    "min_permeability": {
+                        MK.Type: types.Number,
+                        MK.AllowNone: True,
+                        MK.Description: "Minimum allowed permeability in mD before a tube is removed.",
+                    },
```
edubarrosTNO, line 444: Suggestion: `"(...) before a tube is removed (i.e., its cells are made inactive)."`

Plus, would we like to have this to be a different value across e.g. possible different layer zones?

edubarrosTNO (changes requested):
LGTM, just some minor comments

CHANGELOG.md
wouterjdb: I think this is up to the user. You wouldn't set it equal to the minimum from the distribution otherwise you'll never remove anything....

src/flownet/config_parser/_config_parser.py
wouterjdb: > Suggestion: `"(...) before a tube is removed (i.e., its cells are made inactive)."`

Done ✔️ 

> Plus, would we like to have this to be a different value across e.g. possible different layer zones?

Maybe, but I think we should first test whether this has a positive impact before making things more complex.

edubarrosTNO (approved):
👍

PR #351, opened 2021-03-05
Add simple plotting tool

### LINKED ISSUE #148: Speed up plotting
Plotting many realizations is a very slow process. A possible solution could be to:

- Dump to parquet in each realization
- Make a simple command line tool that plots the desired results.

The functionality should be limited - in no way do we want to compete with Webviz/ResInsight etc. We merely need a _pocket knife_ for testing purposes.

wouterjdb:
This PR adds a simple plotting tool which allows you to plot the results of FlowNet runs.

---

### Contributor checklist

- [x] :tada: This PR closes #148 
   - [x] Add initial version of the tool
   - [x] Make more updates
   - [x] Add observations
   - [x] Add support for multiple vertical lines

### DIFF
--- a/src/flownet/utils/plot_results.py
+++ b/src/flownet/utils/plot_results.py
@@ -74,7 +74,7 @@ def plot(
     reference_simulation: Optional[EclSum],
     plot_settings: dict,
 ):
-    """Main plotting function that generate builds up a single plot build up
+    """Main plotting function that generates a single plot
     from potentially multiple ensembles and other data.
 
     Args:
@@ -87,10 +87,10 @@ def plot(
     """
     plt.figure()  # (figsize=[16, 8])
 
-    if len(prior_data):
+    if prior_data:
         plot_ensembles("prior", vector, prior_data, plot_settings)
 
-    if len(posterior_data):
+    if posterior_data:
         plot_ensembles("posterior", vector, posterior_data, plot_settings)
 
     if reference_simulation:
@@ -163,12 +163,13 @@ def check_args(args):
         )
 
     if (
-        not len(args.prior) > 0
-        and not len(args.posterior) > 0
-        and not args.reference_simulation
+        args.prior is None
+        and args.posterior is None
+        and args.reference_simulation is None
     ):
         raise ValueError(
-            "There is no prior, posterior or reference simulation to plot. Supply at least at one of the three to plot."
+            "There is no prior, no posterior and no reference simulation to plot. Supply at least at one "
+            "of the three to plot."
         )
 
     if not (len(args.prior_colors) == 1 or len(args.prior_colors) == len(args.prior)):
@@ -188,7 +189,7 @@ def check_args(args):
 
 
 def build_ensemble_df_list(
-    ensemble_paths: List[str], vectors: List[str]
+    ensemble_paths: Optional[List[str]], vectors: List[str]
 ) -> List[pd.DataFrame]:
     """Helper function to read and prepare ensemble data.
 
@@ -202,19 +203,20 @@ def build_ensemble_df_list(
     """
     data: list = []
 
-    for prior in ensemble_paths:
+    if ensemble_paths is not None:
+        for prior in ensemble_paths:
 
-        df_data = ensemble.ScratchEnsemble(
-            "flownet_ensemble",
-            paths=prior.replace("%d", "*"),
-        ).get_smry(column_keys=vectors)
+            df_data = ensemble.ScratchEnsemble(
+                "flownet_ensemble",
+                paths=prior.replace("%d", "*"),
+            ).get_smry(column_keys=vectors)
 
-        df_data_sorted = df_data.sort_values("DATE")
-        df_realizations = df_data_sorted[
-            df_data_sorted["DATE"] == df_data_sorted.values[-1][0]
-        ]["REAL"]
+            df_data_sorted = df_data.sort_values("DATE")
+            df_realizations = df_data_sorted[
+                df_data_sorted["DATE"] == df_data_sorted.values[-1][0]
+            ]["REAL"]
 
-        data.append(df_data.merge(df_realizations, how="inner"))
+            data.append(df_data.merge(df_realizations, how="inner"))
 
     return data
 
@@ -240,6 +242,7 @@ def main():
         "-prior",
         type=str,
         nargs="+",
+        default=None,
         help="One or more paths to prior ensembles separated by a space. "
         "The path should include a '%d' which indicates the realization number. "
         "Example: runpath/realization-%d/iter-0/",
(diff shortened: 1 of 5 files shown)


src/flownet/utils/plot_results.py
```diff
@@ -0,0 +1,372 @@
...
+        default=[""],
+        nargs="+",
+        help="One or #vectors unit labels.",
+    )
+    parser.add_argument(
+        "-prior_alphas",
+        type=float,
+        default=[0.1],
+        nargs="+",
+        help="One or #prior ensembles alpha (transparency) values.",
```
LonnekevB, line 284:
```suggestion
        help="Transparency of prior, value between 0 (transparent) and 1(opaque). Can be one number or multiple values, depending on the number of priors you are plotting. In the latter case, the number of alpha values should be equal to the number of priors. ",
```

src/flownet/utils/plot_results.py
```diff
@@ -0,0 +1,372 @@
...
+    )
+    parser.add_argument(
+        "-posterior_alphas",
+        type=float,
+        default=[0.1],
+        nargs="+",
+        help="One or #posterior ensembles alpha (transparency) values.",
```
LonnekevB, line 291:
```suggestion
        help="Transparency of posterior, value between 0 (transparent) and 1(opaque). Can be one number or multiple values, depending on the number of posteriors you are plotting. In the latter case, the number of alpha values should be equal to the number of posteriors. ",
```

src/flownet/utils/plot_results.py
```diff
@@ -0,0 +1,372 @@
...
+    )
+    parser.add_argument(
+        "-prior_colors",
+        type=str,
+        default=["gray"],
+        nargs="+",
+        help="One or #prior ensembles colors.",
```
LonnekevB, line 298:
```suggestion
        help="Color of prior lines. Can be one number or multiple colors, depending on the number of priors you are plotting. In the latter case, the number of colors should be equal to the number of priors. ",
```

src/flownet/utils/plot_results.py
```diff
@@ -0,0 +1,372 @@
...
+    )
+    parser.add_argument(
+        "-posterior_colors",
+        type=str,
+        default=["blue"],
+        nargs="+",
+        help="One or #posterior ensembles colors.",
```
LonnekevB, line 305:
```suggestion
        help="Color of posterior. Can be one number or colors, depending on the number of posteriors you are plotting. In the latter case, the number of colors should be equal to the number of posteriors. ",
```

LonnekevB (changes requested):
Just some suggestions on making the help a bit more explicit.

LonnekevB (approved):

src/flownet/utils/plot_results.py
```diff
@@ -0,0 +1,372 @@
...
+        raise ValueError(
+            f"You should either supply a single maximum y-value or as many as you have vectors ({len(args.vectors)}."
+        )
+
+    if not (len(args.units) == 1 or len(args.units) == len(args.vectors)):
+        raise ValueError(
+            f"You should either supply a single units label or as many as you have vectors ({len(args.vectors)}."
+        )
+
+    if (
```
olelod, line 165: If prior or posterior is not defined you will run into trouble here, I think? len(None) is not defined. Better with `if not args.prior and not args.posterior and not args.reference_simulation`?

src/flownet/utils/plot_results.py
```diff
@@ -0,0 +1,372 @@
...
+        default=None,
+        help="Path to an ERT observation file.",
+    )
+    args = parser.parse_args()
+
+    check_args(args)
+
+    prior_data = build_ensemble_df_list(args.prior, args.vectors)
+    posterior_data = build_ensemble_df_list(args.posterior, args.vectors)
+
```
olelod, line 352: Same here. You can end up trying to iterate over `None` inside `build_ensemble_df_list`.
```suggestion
prior_data = build_ensemble_df_list(args.prior, args.vectors) if args.prior else None
posterior_data = build_ensemble_df_list(args.posterior, args.vectors) if args.posterior else None
```

src/flownet/utils/plot_results.py
```diff
@@ -0,0 +1,372 @@
...
+        vector: Name of the vector to plot.
+        prior_data: List of prior ensemble data DataFrames.
+        posterior_data: List of posterior ensemble data DataFrames.
+        reference_simulation: EclSum object for the reference simulation.
+        plot_settings: Settings dictionary for the plots.
+
+    """
+    plt.figure()  # (figsize=[16, 8])
+
+    if len(prior_data):
```
olelod, line 90:
```suggestion
    if prior_data:
```

src/flownet/utils/plot_results.py
```diff
@@ -0,0 +1,372 @@
...
+        plot_ensembles("prior", vector, prior_data, plot_settings)
+
+    if len(posterior_data):
```
olelod, line 93:
```suggestion
    if posterior_data:
```

olelod (changes requested):

src/flownet/utils/plot_results.py
```diff
@@ -0,0 +1,372 @@
...
+
+
+def plot(
+    vector: str,
+    prior_data: list,
+    posterior_data: list,
+    reference_simulation: Optional[EclSum],
+    plot_settings: dict,
+):
+    """Main plotting function that generate builds up a single plot build up
```
olelod, line 77: A bit hard to read... :)

```suggestion
    """Main plotting function that generate a single plot from potentially multiple ensembles and other data
```

or something

wouterjdb:
Example from @edubarrosTNO 

image

src/flownet/utils/plot_results.py
wouterjdb: Fixed by checking in build_ensemble_df_list if the input is None, if so I return an empty list (to make sure I at least return a list which is expected later on).

src/flownet/utils/plot_results.py
wouterjdb: Fixed with 

```
    if (
        not args.prior is None
        and args.posterior is None
        and args.reference_simulation is None
    ):
```

And given all those parameters a default of `None` in argparse.

src/flownet/utils/plot_results.py
wouterjdb: Now always a list and should therefore nicely evaluate to True/False. But your shorter way works just as well, so I have changed it.

src/flownet/utils/plot_results.py
wouterjdb: Now always a list and should therefore nicely evaluate to True/False. But your shorter way works just as well, so I have changed it.

olelod (approved):
LGTM!

PR #372, opened 2021-03-25
Node placement in volume of original model

### LINKED ISSUE #368: Node placement in volume of original model
Flownet now uses the convex hull of the well nodes as outer border for the placement of new nodes. This means that the newly created voronoi volume distribution (#356) does not always work well as it distributes the volume of the original model. Could we add an option that in the node distribution not the convex hull but the volume of the active cells in the (layer of the) original model is used as boundaries for node placement. This will lead to a more accurate representation of the geometry of the original model.

I think this can also still be defended as "data-driven" as one could get this information from the seismic - just as it was originally done.

LonnekevB:

---

### Contributor checklist

- [ ] :tada: This PR closes #368.
   - [ ] Task 1
   - [ ] Task 2

### DIFF
--- a/src/flownet/config_parser/_config_parser.py
+++ b/src/flownet/config_parser/_config_parser.py
@@ -386,7 +386,7 @@ def create_schema(config_folder: Optional[pathlib.Path] = None) -> Dict:
                         MK.Type: types.Bool,
                         MK.AllowNone: True,
                         MK.Description: "When true use boundary of reservoir/layer volume as "
-                        "bounding volume to place initial candidates instead of concave hull of well perforations.",
+                        "bounding volume to place initial candidates instead of convex hull of well perforations.",
                     },
                     "hull_factor": {
                         MK.Type: types.Number,
--- a/src/flownet/config_parser/_config_parser.py
+++ b/src/flownet/config_parser/_config_parser.py
@@ -379,15 +379,14 @@ def create_schema(config_folder: Optional[pathlib.Path] = None) -> Dict:
                     "place_nodes_in_volume_reservoir": {
                         MK.Type: types.Bool,
                         MK.AllowNone: True,
-                        MK.Description: "Use boundary of reservoir volume as "
-                        "bounding box to place additional well nodes",
+                        MK.Description: "Use boundary of reservoir/layer volume as "
+                        "bounding cube to place additional well nodes",
                     },
-                    "hull_factor_well_perforations": {
+                    "hull_factor": {
                         MK.Type: types.Number,
                         MK.Default: 1.2,
-                        MK.Description: "Extend the cube around the well "
-                        "perforations to place additional nodes. "
-                        "Only used when place_nodes_in_reservoir_volume is not True",
+                        MK.Description: "Increase the bounding cube around the well "
+                        "perforations to place additional nodes.",
                     },
                     "random_seed": {
                         MK.Type: types.Number,
--- a/src/flownet/config_parser/_config_parser.py
+++ b/src/flownet/config_parser/_config_parser.py
@@ -376,7 +376,19 @@ def create_schema(config_folder: Optional[pathlib.Path] = None) -> Dict:
                         MK.Description: "Number of additional nodes to create "
                         "(using Mitchell's best candidate algorithm)",
                     },
-                    "hull_factor": {MK.Type: types.Number, MK.Default: 1.2},
+                    "place_nodes_in_volume_reservoir": {
+                        MK.Type: types.Bool,
+                        MK.AllowNone: True,
+                        MK.Description: "Use boundary of reservoir volume as "
+                        "bounding box to place additional well nodes",
+                    },
+                    "hull_factor_well_perforations": {
+                        MK.Type: types.Number,
+                        MK.Default: 1.2,
+                        MK.Description: "Extend the cube around the well "
+                        "perforations to place additional nodes. "
+                        "Only used when place_nodes_in_reservoir_volume is not True",
+                    },
                     "random_seed": {
                         MK.Type: types.Number,
                         MK.AllowNone: True,
--- a/src/flownet/config_parser/_config_parser.py
+++ b/src/flownet/config_parser/_config_parser.py
@@ -276,7 +276,7 @@ def create_schema(config_folder: Optional[pathlib.Path] = None) -> Dict:
                                 MK.Type: types.Bool,
                                 MK.AllowNone: True,
                                 MK.Description: "When true, the bounding boxes of the gridcells of the "
-                                "original reservoir model are used check if the generated additional "
+                                "original reservoir model are used to check if the generated additional "
                                 "nodes are positioned within the reservoir volume.",
                             },
                         },
@@ -386,12 +386,12 @@ def create_schema(config_folder: Optional[pathlib.Path] = None) -> Dict:
                         MK.Type: types.Bool,
                         MK.AllowNone: True,
                         MK.Description: "When true use boundary of reservoir/layer volume as "
-                        "bounding cube to place initial candidates instead of concave hull of well perforations",
+                        "bounding volume to place initial candidates instead of concave hull of well perforations.",
                     },
                     "hull_factor": {
                         MK.Type: types.Number,
                         MK.Default: 1.2,
-                        MK.Description: "Increase the size of the bounding cube around the well "
+                        MK.Description: "Increase the size of the bounding volume around the well "
                         "perforations to place additional nodes in.",
                     },
                     "random_seed": {
@@ -1699,7 +1699,7 @@ def parse_config(
     ):
         raise ValueError(
             "concave_hull needs to be true for flownet to be able to "
-            "place candidates within the reservoir volume"
+            "place candidates within the reservoir volume."
         )
 
     req_relp_parameters: List[str] = []
--- a/src/flownet/config_parser/_config_parser.py
+++ b/src/flownet/config_parser/_config_parser.py
@@ -272,7 +272,13 @@ def create_schema(config_folder: Optional[pathlib.Path] = None) -> Dict:
                                 "are supported,  e.g. weekly (W), monthly (M), quarterly (Q), "
                                 "yearly (A)",
                             },
-                            "concave_hull": {MK.Type: types.Bool, MK.AllowNone: True},
+                            "concave_hull": {
+                                MK.Type: types.Bool,
+                                MK.AllowNone: True,
+                                MK.Description: "When true, the bounding boxes of the gridcells of the "
+                                "original reservoir model are used check if the generated additional "
+                                "nodes are positioned within the reservoir volume.",
+                            },
                         },
                     },
                     "constraining": {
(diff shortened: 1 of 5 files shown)


wouterjdb (commented):
I tested the code - but had to do some minor such that the random point generation actually happens in the space of the original model.

This is a picture of the Norne model (top layer). Now stretching all the way to the west - the area that doesn't have any wells.

image

The full Norne model, with many nodes per layer, would now look like:

image

And for reference the real Norne model (initial pressure is quite off by the way, @edubarrosTNO I hope this is not the case in the cases you have been testing?):

image

edubarrosTNO:
> And for reference the real Norne model (initial pressure is quite off by the way, @edubarrosTNO I hope this is not the case in the cases you have been testing?):

I will have to check that, but what I can say now is that those cases that I have been running are certainly using much fewer additional nodes than what we see in these pictures.
Are there any current hypotheses for the initial pressure being off? To me it looks like the pressure is in the right range in most of the model, except for those red spots in one of the "tentacles". It would be useful to check the EQUIL regions of the full grid and FlowNet models, and see if there is any overlapping with the red spots in pressure in the FlowNet grid. If that is the case, we will have to revise the EQUIL settings of that region

src/flownet/config_parser/_config_parser.py
```diff
@@ -376,7 +376,18 @@ def _to_abs_path(path: Optional[str]) -> str:
                         MK.Description: "Number of additional nodes to create "
                         "(using Mitchell's best candidate algorithm)",
                     },
-                    "hull_factor": {MK.Type: types.Number, MK.Default: 1.2},
+                    "place_nodes_in_volume_reservoir": {
+                        MK.Type: types.Bool,
+                        MK.AllowNone: True,
```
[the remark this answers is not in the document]
LonnekevB: This just allows to not actually have the place_nodes_in_volume_reservoir keyword in the yaml. If it is not defined it is automatically set to None.

src/flownet/network_model/_mitchell.py
```diff
@@ -55,46 +58,59 @@ def mitchell_best_candidate_modified_3d(
...
+    if place_nodes_in_volume_reservoir:
+        x_mins, x_maxs, y_mins, y_maxs, z_mins, z_maxs = (
+            np.asarray(t) for t in zip(*concave_hull_bounding_boxes)
+        )
+        x_min = min(x_mins + x_maxs)
+        x_max = max(x_mins + x_maxs)
+        y_min = min(y_mins + y_maxs)
+        y_max = max(y_mins + y_maxs)
+        z_min = min(z_mins + z_maxs)
+        x_max = max(z_mins + z_maxs)
```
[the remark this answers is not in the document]
LonnekevB: Good catch!

src/flownet/config_parser/_config_parser.py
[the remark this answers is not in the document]
wouterjdb: Woops - my mistake. I read that the default was set to `True`, but we're talking about `Mk.AllowNone`, that is something else.

src/flownet/config_parser/_config_parser.py
```diff
@@ -272,7 +272,13 @@ def _to_abs_path(path: Optional[str]) -> str:
                                 "are supported,  e.g. weekly (W), monthly (M), quarterly (Q), "
                                 "yearly (A)",
                             },
-                            "concave_hull": {MK.Type: types.Bool, MK.AllowNone: True},
+                            "concave_hull": {
+                                MK.Type: types.Bool,
+                                MK.AllowNone: True,
+                                MK.Description: "When true, the bounding boxes of the gridcells of the "
+                                "original reservoir model are used check if the generated additional "
```
wouterjdb, line 279:
```suggestion
                                "original reservoir model are used to check if the generated additional "
```

src/flownet/config_parser/_config_parser.py
```diff
@@ -376,7 +382,18 @@ def _to_abs_path(path: Optional[str]) -> str:
                         MK.Description: "Number of additional nodes to create "
                         "(using Mitchell's best candidate algorithm)",
                     },
-                    "hull_factor": {MK.Type: types.Number, MK.Default: 1.2},
+                    "place_nodes_in_volume_reservoir": {
+                        MK.Type: types.Bool,
+                        MK.AllowNone: True,
+                        MK.Description: "When true use boundary of reservoir/layer volume as "
+                        "bounding cube to place initial candidates instead of concave hull of well perforations",
```
wouterjdb, line 389: The individual cells are "cube-like", but the sum of the cubes not. So maybe this fits better?

```suggestion
                        "bounding volume to place initial candidates instead of concave hull of well perforations.",
```

src/flownet/config_parser/_config_parser.py
```diff
@@ -376,7 +382,18 @@ def _to_abs_path(path: Optional[str]) -> str:
...
+                    },
+                    "hull_factor": {
+                        MK.Type: types.Number,
+                        MK.Default: 1.2,
+                        MK.Description: "Increase the size of the bounding cube around the well "
```
wouterjdb, line 394:
```suggestion
                        MK.Description: "Increase the size of the bounding volume around the well "
```

src/flownet/config_parser/_config_parser.py
```diff
@@ -1676,6 +1693,15 @@ def parse_config(
             "there is only a single layer."
         )
 
+    if (
+        config.flownet.place_nodes_in_volume_reservoir
+        and not config.flownet.data_source.concave_hull
+    ):
+        raise ValueError(
+            "concave_hull needs to be true for flownet to be able to "
+            "place candidates within the reservoir volume"
```
wouterjdb, line 1702:
```suggestion
            "place candidates within the reservoir volume."
```

src/flownet/network_model/_mitchell.py
```diff
@@ -175,3 +183,41 @@ def mitchell_best_candidate_modified_3d(
 
     # Return the real/original and added flow node coordinates as a list of tuples.
     return [(x[i], y[i], z[i]) for i in range(len(x))]
+
+
+def scale_concave_hull_perforations(
+    perforations: List[Coordinate], hull_factor: float
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+    """
+
```
wouterjdb, line 192: Missing docstring.

src/flownet/network_model/_mitchell.py
```diff
@@ -175,3 +183,41 @@ def mitchell_best_candidate_modified_3d(
...
+        perforations: Python list of real well coordinate tuples
+            [(xr_1, yr_1, zr_1), ..., (xr_N, yr_N, zr_N)]
+        hull_factor: Factor to linearly scale the convex hull with. Factor will
+            scale the distance of each point from the centroid of all the points.
+    Returns:
+        The tuple consisting of numpy arrays of x,y,z moved points based on the hull_factor,
+        which are used further on in the code to create a concave hull around the real wells.
+
+    """
+
```
wouterjdb, line 204:
```suggestion
```

src/flownet/network_model/_mitchell.py
```diff
@@ -12,7 +12,8 @@ def mitchell_best_candidate_modified_3d(
     perforations: List[Coordinate],
     num_added_flow_nodes: int,
     num_candidates: int = 1000,
-    hull_factor: Optional[float] = None,
+    place_nodes_in_volume_reservoir: Optional[bool] = None,
+    hull_factor: float = 1.2,
```
wouterjdb, line 16: Why 1.2? Previously it was `None`.

wouterjdb (changes requested):

src/flownet/network_model/_mitchell.py
LonnekevB: @wouterjdb : True, but it is defaulted in the config parser to 1.2, similarly as `num_candidates` is also defaulted to 1000. So that's why I aligned the notation, as it was never `None` anyway. But maybe as they always have a value anyway, we should just not re-assign them here at all (like `num_add_flow_nodes`)

LonnekevB:
@wouterjdb  I realized we were using convex and concave hull interchangeably in the Mitchells code. But for the grid cell bounding boxes it is a concave hull, but for the perforations, we actually use a convex hull. 
I think the naming is consistent now and once the tests have passed ready to merge!

src/flownet/network_model/_mitchell.py
```diff
@@ -55,45 +58,48 @@ def mitchell_best_candidate_modified_3d(
...
-        z_max = max(z)
-
-    # Determine the convex hull of the original or linearly scaled perforations
-    if np.all(z == z[0]):
-        # 2D cases
-        hull = Delaunay(np.column_stack([x, y]))
-    else:
-        # 3D cases
-        hull = Delaunay(np.column_stack([x, y, z]))
+        # Determine whether the complex hull needs to be scaled
```
olelod:
```suggestion
        # Determine whether the convex hull needs to be scaled
```

src/flownet/network_model/_mitchell.py
```diff
@@ -55,45 +58,48 @@ def mitchell_best_candidate_modified_3d(
...
+        if not np.isclose(hull_factor, 1.0):
+            x_hull, y_hull, z_hull = scale_concave_hull_perforations(
```
olelod:
```suggestion
            x_hull, y_hull, z_hull = scale_convex_hull_perforations(
```

src/flownet/network_model/_mitchell.py
```diff
@@ -175,3 +183,41 @@ def mitchell_best_candidate_modified_3d(
 
     # Return the real/original and added flow node coordinates as a list of tuples.
     return [(x[i], y[i], z[i]) for i in range(len(x))]
+
+
+def scale_concave_hull_perforations(
```
olelod:
```suggestion
def scale_convex_hull_perforations(
```

src/flownet/network_model/_mitchell.py
```diff
@@ -175,3 +183,41 @@ def mitchell_best_candidate_modified_3d(
...
+
+
+    Args:
+        perforations: Python list of real well coordinate tuples
+            [(xr_1, yr_1, zr_1), ..., (xr_N, yr_N, zr_N)]
+        hull_factor: Factor to linearly scale the convex hull with. Factor will
+            scale the distance of each point from the centroid of all the points.
+    Returns:
+        The tuple consisting of numpy arrays of x,y,z moved points based on the hull_factor,
+        which are used further on in the code to create a concave hull around the real wells.
```
olelod:
```suggestion
        which are used further on in the code to create a convex hull around the real wells.
```

olelod (approved):
I see that during the time I did something else, all of my comments about convex/concave became outdated...

Just one comment left.

LGTM! Nice work.

olelod (changes requested):
I just have one comment about a docstring mentioning the `hull_factor = 1.0`. Other than that everything looks good to go! Nice job.

olelod (approved):

wouterjdb (approved):

PR #379, opened 2021-04-21
Add delta to SWL to calculate SWCR

olelod:
If you want to create a wide range for both SWL and SWCR there is a large chance for generating non-physical values, where SWCR is lower than SWL. This PR adds a parameter to the config yaml (defaulted to false), that provides the opportunity to calculate the SWCR from the SWL by adding a delta value. 

If `swcr_add_to_swcr` is set to `True` the `swcr` parameter in the config yaml will be treated as a prior distribution for the delta value that will be added to SWL, not as a prior distribution for SWCR directly.

---

### Contributor checklist

   - [x] Add `swcr_add_to_swl` to config_parser
   - [x] Update render_output for relative_permeability

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## Unreleased
 
 ### Added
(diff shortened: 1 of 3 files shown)


wouterjdb (approved):
LGTM 🚀

PR #380, opened 2021-04-22
I377 Mitchell's speed up

### LINKED ISSUE #377: Speed up Mitchells best candidate algorithm
A start has been made by @wouterjdb, branch: https://github.com/wouterjdb/flownet/tree/mitchell-speed-up

LonnekevB:

---

### Contributor checklist

- [ ] :tada: This PR closes #377.
   - [ ] Task 1
   - [ ] Task 2


LonnekevB:
I looked through the code and I think it is all mathematically correct what we are doing in the fast Mitchells algorithm. I plotted the flownodes from both algorithms (Norne single layer, 20 additional flow nodes, 100 candidates, just a quick meaningless check). I will see if I can plot the actual wells in a different color, to make it a bit more clear.
image


The fast Mitchell algorithm is now also included in the unit tests and they pass. So additional nodes are placed correctly in the volume of the reservoir at least.

wouterjdb:
Looking at the images, I'm wondering about the depths. In the "normal" version there are a lot more very deep (shallow in the picture?) nodes. Why?

olelod:
> Looking at the images, I'm wondering about the depths. In the "normal" version there are a lot more very deep (shallow in the picture?) nodes. Why?

I think the difference in z-scale is making it a bit difficult to compare here @wouterjdb.

wouterjdb (changes requested):
Some minor comments.

I tried running for Norne with hundred of nodes to be added and 5,000/50,000 candidates in "Fast" mode. It runs nice 'n fast, but unfortunately I get into the following error:

image

If I run in "Normal" mode I don't get these errors.

src/flownet/network_model/_mitchell_fast.py
```diff
@@ -0,0 +1,179 @@
...
+        mid = time.time()
+        if mid - start > 4:
+            start = mid
+            print(
+                f"\rAdding flow nodes:  {int(((i-num_points)/num_added_flow_nodes)*100)}%",
+                end="",
+            )
+
+        delta_x_relative = np.power(
+            (x.repeat(num_candidates) - x_candidate.repeat(x.shape[0]))
```
olelod, line 148: using `repeat `is actually quite neat, I think, but in order to get all combinations correct I think you need to use `tile` also
```suggestion
            (np.tile(x, num_candidates) - x_candidate.repeat(x.shape[0]))
```

src/flownet/network_model/_mitchell_fast.py
```diff
@@ -0,0 +1,179 @@
...
+            / (x_max - x_min),
+            2,
+        )
+        delta_y_relative = np.power(
+            (y.repeat(num_candidates) - y_candidate.repeat(y.shape[0]))
```
olelod, line 153:
```suggestion
            (np.tile(y, num_candidates) - y_candidate.repeat(y.shape[0]))
```

src/flownet/network_model/_mitchell_fast.py
```diff
@@ -0,0 +1,179 @@
...
+            / (y_max - y_min),
+            2,
+        )
+        if np.all(z == z[0]):
+            delta_z_relative = 0
+        else:
+            delta_z_relative = np.power(
+                (
+                    z.repeat(num_candidates)
```
olelod, line 162:
```suggestion
                    np.tile(z, num_candidates)
```

src/flownet/network_model/_mitchell_fast.py
```diff
@@ -0,0 +1,179 @@
...
+                    - z_candidate.repeat(z.shape[0]) / (z_max - z_min)
+                ),
+                2,
+            )
+
+        dists = np.sqrt(delta_x_relative + delta_y_relative + delta_z_relative)
+        best_candidate = np.argmax(dists)
```
olelod, line 169: Now you are not looking for the longest distance, but the "longest shortest distance". For each candidate point, find the shortest distance any points already included. We are after the longest of these shortest distances.
```suggestion
        best_candidate = np.where(dists == np.max(np.min(np.reshape(dists, (len(x), num_candidates)), axis=1)))
```

src/flownet/network_model/_mitchell_fast.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # Add the best candidate's coordinates; a new flow node is added
+        x = np.append(x, x_candidate.repeat(x.shape[0])[best_candidate])
+        y = np.append(y, y_candidate.repeat(y.shape[0])[best_candidate])
+        z = np.append(z, z_candidate.repeat(z.shape[0])[best_candidate])
```
olelod, line 174: Here I guess we always end up with size of ` x, y, z` increasing while `num_candidates` remain the same, so after the first point is added we will always propose points that are already in the model. Since the shortest distance will be 0 to all these points it will never want to be added again - so it should not be a problem. An alternative would be to remove the point from the candidates, but I don't  think it is necessary.

olelod (changes requested):

olelod:
I also get a feeling that 

```
        in_hull = np.asarray([False] * num_candidates)
        x_candidate = np.zeros(num_candidates)
        y_candidate = np.zeros(num_candidates)
        z_candidate = np.zeros(num_candidates)

        # Repeat while not all random points are inside the convex hull
        while not all(in_hull):
            # Generate a set of random candidates that will be the new
            # flow nodes
            x_candidate_tmp = x_min + np.random.rand(num_candidates) * (x_max - x_min)
            y_candidate_tmp = y_min + np.random.rand(num_candidates) * (y_max - y_min)
            z_candidate_tmp = z_min + np.random.rand(num_candidates) * (z_max - z_min)

            # Update the list of flow node candidates. Only the points that previously
            # were not inside the convex hull are updated.
            np.putmask(x_candidate, np.invert(in_hull), x_candidate_tmp)
            np.putmask(y_candidate, np.invert(in_hull), y_candidate_tmp)
            np.putmask(z_candidate, np.invert(in_hull), z_candidate_tmp)

            candidates = np.vstack([x_candidate, y_candidate, z_candidate]).T

            if concave_hull_bounding_boxes is not None:
                # Test whether all points are inside the bounding boxes of the gridcells of the reservoir volume.
                # This is the always the case when place_nodes_in_volume_reservoir is True.
                in_hull = check_in_hull(
                    concave_hull_bounding_boxes, candidates, in_hull_known=in_hull
                )
            else:
                # Test whether all points are inside the convex hull of the perforations
                if np.all(z == z[0]):
                    in_hull = perforation_hull.find_simplex(candidates[:, (0, 1)]) >= 0
                else:
                    in_hull = perforation_hull.find_simplex(candidates) >= 0
```

could have been made as a `_generate_candidates` function. This could be called on the inside or outside of the for-loop like so:

```
    if fast_option:
         in_hull, x_candidates, y_candidates, z_candidates = _generate_candiadates( some arguments )

    # Generate all new flow nodes
    for i in range(num_points, num_points + num_added_flow_nodes):
        mid = time.time()
        if mid - start > 4:
            start = mid
            print(
                f"\rAdding flow nodes:  {int(((i-num_points)/num_added_flow_nodes)*100)}%",
                end="",
            )
        if not fast_option:
            in_hull, x_candidates, y_candidates, z_candidates = _generate_candiadates( some arguments )

```

Or are there more differences between `mitchells_best_candidate` and `mitchells_best_candidate_fast` that I am not picking up?

LonnekevB:
> Some minor comments.
> 
> I tried running for Norne with hundred of nodes to be added and 5,000/50,000 candidates in "Fast" mode. It runs nice 'n fast, but unfortunately I get into the following error:
> 
> image
> 
> If I run in "Normal" mode I don't get these errors.

I ran it with 100 nodes and 50.000 candidates in fast mode and it runs fine. So this issue seems to be fixed :)

src/flownet/network_model/_mitchell.py
```diff
@@ -94,13 +93,26 @@ def mitchell_best_candidate_modified_3d(
...
+            num_candidates,
+            x_min,
+            x_max,
+            y_min,
+            y_max,
+            z_min,
+            z_max,
+            perforation_hull,
+            concave_hull_bounding_boxes,
+        )
```
wouterjdb, line 114: You are writing this list of arguments twice in the code. To save some space you could write:

```
arguments = { "arg1": 1, "arg2": 2}
function(**arguments)
```

src/flownet/network_model/_mitchell.py
```diff
@@ -111,88 +123,132 @@ def mitchell_best_candidate_modified_3d(
...
+        else:
+            if perforation_hull is not None:
+                # Test whether all points are inside the convex hull of the perforations
+                if np.isclose(z_min, z_max):
+                    in_hull = perforation_hull.find_simplex(candidates[:, (0, 1)]) >= 0
+                else:
+                    in_hull = perforation_hull.find_simplex(candidates) >= 0
+            else:
+                raise ValueError(
+                    "perforation_hull should never be None when concave_hull_bounding_boxes is also None"
```
wouterjdb, line 239:
```suggestion
                    "The perforation_hull should never be None when concave_hull_bounding_boxes is also None."
```

Can this actually happen at this point?

wouterjdb (changes requested):
Minor comments....

I will also try to run the Norne case with the fast mode, but Azure seems to have some issues this morning....

src/flownet/network_model/_mitchell.py
LonnekevB: No I am pretty sure it can't, but pylint still complained. 

`src/flownet/network_model/_mitchell.py:233: error: Item "None" of "Optional[Any]" has no attribute "find_simplex"
src/flownet/network_model/_mitchell.py:235: error: Item "None" of "Optional[Any]" has no attribute "find_simplex"
Found 2 errors in 1 file (checked 61 source files)
Error: Process completed with exit code 1.`

Didn't know how else to fix this.

olelod: Can you do like this? 
image

wouterjdb:
Tested running with fast mode (5000 candidates). It took only a handful of seconds to generate the grid, and it looked like:

image

Seems ok 👍

src/flownet/network_model/_mitchell.py
LonnekevB: Yes this works actually. I got a bit confused because concave_hull_bounding_boxes and perforation_hull can both be true and if that is the case you only want the `concave_hull_bounding_boxes is not None` statement to run. But since that one is the first if statement, that should be fine indeed.

src/flownet/network_model/_mitchell.py
LonnekevB: I could do this. But then the function itself becomes a bit unclear with just **arguments. I would prefer to keep it this way to have the function complete with all the keywords listed.

wouterjdb (approved):
LGTM 👍

PR #383, opened 2021-04-23
Add delta to KRWEND to calculate KRWMAX

olelod:
If you want to create a wide range for both KRWMAX and KRWEND there is a large chance for generating non-physical values, where KRWEND is lower than KRWMX. This PR adds a parameter to the config yaml (defaulted to false), that provides the opportunity to calculate the KRWMAX from the KRWEND by adding a delta value.

If krwmax_add_to_krwend is set to True the krwend parameter in the config yaml will be treated as a prior distribution for the delta value that will be added to KRWEND, not as a prior distribution for KRWMAX directly.

The PR also adds the possibility to define a prior for KRWMAX directly. This value was previously defaulted to 1 only.

---

### Contributor checklist

   - [ ] Add krwmax_add_to_krwend to config_parser
   - [ ] Add krwmax to config_parser
   - [ ] Add checks in config_parser
   - [ ] Add krwmax to render_realization


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## Unreleased
 
 ### Added
+- [#383](https://github.com/equinor/flownet/pull/383) Added option to define either define a prior distribution for KRWMAX directly by using krwmax in the config yaml, or to let KRWMAX be calculated as KRWEND + delta. To do the latter, set krwmax_add_to_krwend to true, and then the prior distribution definition in the config yaml for krwmax will be interpreted as a prior distribution for the delta value to be added to KRWEND to get the KRWMAX.
```
wouterjdb, line 8:
```suggestion
- [#383](https://github.com/equinor/flownet/pull/383) Added option to either define a prior distribution for KRWMAX directly by using krwmax in the config yaml, or to let KRWMAX be calculated as KRWEND + delta. To do the latter, set krwmax_add_to_krwend to true, and then the prior distribution definition in the config yaml for krwmax will be interpreted as a prior distribution for the delta value to be added to KRWEND to get the KRWMAX.
```

src/flownet/parameters/_relative_permeability.py
```diff
@@ -221,6 +226,30 @@ def __init__(
...
+            Updated dictionary with parameters
+        """
+        if "krwmax" not in params.keys():
+            params["krwmax"] = 1
+        if self._krwmax_add_to_krwend:
+            params["krwmax"] = min(1, params["krwend"] + params["krwmax"])
+        if params["krwmax"] < params["krwend"]:
+            params["krwmax"] = params["krwend"]
+            warnings.warn(
+                "KRWMAX < KRWEND in one realization. KRWMAX set equal to KRWEND."
```
wouterjdb, line 247: One or more?

CHANGELOG.md
olelod: Do you never define either define something?

src/flownet/parameters/_relative_permeability.py
olelod: Yes

wouterjdb (approved):
LGTM 🤖

PR #387, opened 2021-04-28
Read FlowNet from restart folder pickle files when using restart option

### LINKED ISSUE #198: Use the previously constructed flownet when updating distributions
When restarting a history match and prescribing the mean from the posterior distribution of a previously completed history match, it is not necessary to reconstruct the flownet. Instead the flownet can be read from file as is also done when running a prediction.

olelod:
There is no need to construct the FlowNet when using the option to define prior distributions from a previously completed run. This PR checks if `network/schedule/parameters.pickle` exists, in addition to the `parameter_iteration-latest.parquet.gzip` file. If the files exists, they are read and a new ERT setup is made based on those.

---

### Contributor checklist

- [ ] :tada: This PR closes #198 .
   - [ ] Add `run_flownet_history_matching_from_restart `function to `_run_ahm.py`
   - [ ] Add checks in `_command_line.py` to see if `restart_folder `points to an existing case

### DIFF
--- a/src/flownet/_command_line.py
+++ b/src/flownet/_command_line.py
@@ -80,6 +80,10 @@ def flownet_ahm(args: argparse.Namespace) -> None:
                 ).is_file()
             ):
                 run_flownet_history_matching_from_restart(config, args)
+            else:
+                raise ValueError(
+                    f"Case in {args.restart_folder} is not complete! Some files are missing!"
+                )
         else:
             raise ValueError(f"{args.restart_folder} does not exist!")
 
--- a/src/flownet/_command_line.py
+++ b/src/flownet/_command_line.py
@@ -5,7 +5,7 @@ import pathlib
 import subprocess
 
 from .config_parser import parse_config, parse_pred_config, parse_hyperparam_config
-from .ahm import run_flownet_history_matching
+from .ahm import run_flownet_history_matching, run_flownet_history_matching_from_restart
 from .prediction import run_flownet_prediction
 from .hyperparameter import run_flownet_hyperparameter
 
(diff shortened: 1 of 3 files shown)


src/flownet/_command_line.py
```diff
@@ -66,7 +66,26 @@ def flownet_ahm(args: argparse.Namespace) -> None:
...
+                and pathlib.Path(args.restart_folder / "schedule.pickled").is_file()
+                and pathlib.Path(args.restart_folder / "parameters.pickled").is_file()
+                and pathlib.Path(
+                    args.restart_folder / "parameters_iteration-latest.parquet.gzip"
+                ).is_file()
+            ):
+                run_flownet_history_matching_from_restart(config, args)
+            else:
+                raise ValueError(
+                    f"Case in {args.restart_folder} is not complete! Some files are missing!"
```
wouterjdb, line 83: Maybe indicate which files are needed to thew users - makes it slightly easier to understand the problem.

src/flownet/ahm/_run_ahm.py
```diff
@@ -366,6 +367,47 @@ def update_distribution(
...
+def run_flownet_history_matching_from_restart(
+    config: ConfigSuite.snapshot, args: argparse.Namespace
+):
+    """
+    Creates and runs an ERT setup, given user configuration.
+
+    Args:
+        config: Configsuite parsed user provided configuration.
+        args: Argparse parsed command line arguments.
+    Returns:
```
wouterjdb, line 379:
```suggestion

    Returns:
```

src/flownet/ahm/_run_ahm.py
```diff
@@ -366,6 +367,47 @@ def update_distribution(
...
+    )
+
+    ahm.create_ert_setup(
+        args=args,
+        training_set_fraction=_find_training_set_fraction(schedule, config),
+    )
+
+    ahm.report()
+
+    ahm.run_ert(weights=config.ert.ensemble_weights)
```
wouterjdb, line 408: This is duplicated code - is there a logical way to not duplicate this code?

wouterjdb (commented):
Some minor comments.

src/flownet/_command_line.py
olelod: Agree. Added the file names.

wouterjdb (approved):
LGTM :rock

PR #396, opened 2021-05-04
Make EQLNUM regions in FlowNet based on region parameter from simulation model

olelod:
When setting the scheme to regions_from_sim the EQLNUM regions in the input simulation model were used to generate the EQLNUM regions for the FlowNet model. This PR makes this more flexible by allowing any region parameter from the INIT file to be used.

---

### Contributor checklist

   - [x] Add region_parameter_from_sim_model to config_parser.py for equil
   - [x] Add checks to config_parser.py
   - [x] Change the generation of EQLNUM regions for FLOWNET to use the `region_parameter_from_sim_model` instead of the previously hard coded EQLNUM

### DIFF
--- a/src/flownet/config_parser/_config_parser.py
+++ b/src/flownet/config_parser/_config_parser.py
@@ -1440,6 +1440,12 @@ def create_schema(config_folder: Optional[pathlib.Path] = None) -> Dict:
                                 MK.Default: "global",
                                 MK.Transformation: _to_lower,
                             },
+                            "region_parameter_from_sim_model": {
+                                MK.Type: types.String,
+                                MK.Description: "The name of the regions parameter in the simulation model to "
+                                "base the equilibrium parameter on.",
+                                MK.Default: "EQLNUM",
+                            },
                             "regions": {
                                 MK.Type: types.List,
                                 MK.Content: {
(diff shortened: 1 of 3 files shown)


src/flownet/config_parser/_config_parser.py
```diff
@@ -1757,39 +1754,42 @@ def parse_config(
...
+            _check_if_all_region_priors_defined(
+                getattr(config.model_parameters, reg_param),
+                unique_regions[reg_param],
+                flow_region_name,
+            )
+        else:
+            regions = getattr(getattr(config.model_parameters, reg_param), "regions")
+            if regions[0].id is not None:
+                raise ValueError(
+                    f"The region number for the first {reg_param} region parameter should not be set, \n"
```
wouterjdb, line 1790: This looks like the exact same error as for the relative permeability.

Potential for re-use of code?
Maybe the user should know for which parameter the error is given?

olelod: This is inside a for loop where i think i have done what you ask for
```
    region_parameters: dict = {"equil": "EQLNUM", "relative_permeability": "SATNUM"}
    unique_regions: dict = {}
    for reg_param, flow_region_name in region_parameters.items():
```

wouterjdb: reg_param is "EQLNUM"/"SATNUM" isn't it...? I was thinking to update the warning such that it says:

"For the **relative permeability paramter**, the region number for the first [...]"

We should know that there isn't it? For example the key in the dict should have that info.

olelod: reg_param is "equil" or "relative_permeability" (the keywords in the config yaml). I could of course make another dictionary that relates "equil"/"relative_permeability" to "equilibrium region"/"relative permeability".

wouterjdb: No no, that was not the point. "equil" and "relative_permeability" is ✔️ 

But you are not printing that in the warning, isn't it? Or are you?

wouterjdb: Never mind.

wouterjdb (approved):

PR #401, opened 2021-05-07
Distribute model volume to nearest tube cell midpoint

### LINKED ISSUE #395: Too much volume assigned to (short?) tubes
When volume is assigned to the nearest tube using `voronoi_per_tube` some tubes get assigned much larger volumes then others:
image

In this case roughly 16 times more volume is assigned to this tube compared to the average. Whereas in the original model there is no high-volume zone there, neither is this tube in a boundary region that gets assigned volume from a large area.

Looking at the picture it seems to happen mostly for shorter tubes. What is causing this?

wouterjdb:
This PR changes the way volume from an original simulation model is distributed to FlowNet. Instead of distributing to the mid point of all tubes the volume is distributed to the midpoint of all tube cell mid points.

The volumes distributed to flownet tube cells are then summed per tube, and evenly redistributed over the single tube.

---

### Contributor checklist

- [x] :tada: This PR closes #395.
   - [x] Change algorithm
   - [ ] Test

### DIFF
--- a/src/flownet/data/from_flow.py
+++ b/src/flownet/data/from_flow.py
@@ -390,7 +390,8 @@ class FlowData(FromSource):
         """Generate bulk volume distribution per grid cell in the FlowNet model based on the geometrical
         distribution of the volume in the original (full field) simulation model. I.e., the original model's
         volume will be distributed over the FlowNet's tubes by assigning original model grid cell
-        volumes to the nearest FlowNet tube midpoint.
+        volumes to the nearest FlowNet tube cell midpoint. Finally, the volume distributed to all cells in a tube
+        will be summed and evenly redistributed over the tube.
 
         Args:
             network: FlowNet network instance.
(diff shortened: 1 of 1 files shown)


wouterjdb:
Comparison of volumes (identical color scales):

**Old method**
image

**New method**
image

**Observations:**
The total distributed volume is similar (6.3e8 vs 6.5e8); minor differences are to be expected due to differences is porosity and bulk volume multipliers being different. Using the new distribution scheme there are no extreme outliers anymore for short vertical connections.

The following plot shows the volume distribution using the new algorithm, but now with a color scale that matches better with the assigned volumes.

image

Some outliers can still be observed, but the outlier that has the highest volume per cell has less than 40% of the volume that was previously assigned to the largest grid cell - and there are fewer outliers. This suggest that all-in-all the volume is more evenly distributed (as in, a better approximation of the original model).

Looking at the statistics, the new method has a similar P10, mean and P90 values for the distribution. However, again, the maximum values has significantly reduces - the outliers are reduces.

image

olelod (approved):
LGTM

PR #402, opened 2021-05-10
Fixed THPRES, now 2-way

### LINKED ISSUE #399: Need to provide two-way THPRES values
Currently in FlowNet THPRES values between two adjacent flow tubes are defined only in one direction.
For example:
```
THPRES
787 793 1* /
422 143 1* /
773 895 1* /
768 551 1* /
(...)
```
Is the two-way effect taken into account by default in OPM-Flow? Is it necessary to enable any options in the RUNSPEC section to activate that? Do we need to explicitly add the two-way values in the THPRES keyword? This seems to be the case in the Norne reference simulation case where THPRES is defined both ways with non-defaulted values even though the values are identical, see below.
```
THPRES
  1 2 0.588031 /
  2 1 0.588031 /
  1 3 0.787619 /
  3 1 0.787619 /
  1 4 7.00083  /
  4 1 7.00083  /
/
```

wouterjdb:
This PR updates the THPRES template, such that THPRES is calulcated for flow in both directions.

---

### Contributor checklist

- [x] :tada: This PR closes #399
   - [x] Fix template

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#372](https://github.com/equinor/flownet/pull/372) Added option to let the additional flownodes initially be placed within the original volume rather than within the convex hull of the real wells. To do this set place_nodes_in_volume_reservoir to true.
 
 ### Fixes
+- [#402](https://github.com/equinor/flownet/pull/402) THPRESS is now calculated in both directions of the equil region boundary.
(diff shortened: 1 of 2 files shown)


wouterjdb:
The OPM Manual states that this is only important if the EQLOPTS keyword has the IRREVER option set. Something which is per today not even supported in Flow.

That means that this is a non-issue and doesn't require a fix. Closing the issue and the PR.

PR #403, opened 2021-05-10
Only print THPRES between neighboring equil regions

### LINKED ISSUE #400: Assign THPRES values per region of the reference model
Currently in FlowNet models THPRES values are assigned between each of the adjacent tubes assumed to have each their own equilibration region (EQLNUM). The EQUIL keyword is used based on the equilibration regions of the reference model. Would it be useful to assign THPRES values according to regions of the reference model as well? Could this help improve results?

For example, in the Norne case, the reference model has:
```
-- EQLNUM EQLNUM THPRES
-- FROM TO VALUE
THPRES
  1 2 0.588031 /
  2 1 0.588031 /
  1 3 0.787619 /
  3 1 0.787619 /
  1 4 7.00083  /
  4 1 7.00083  /
/
```

olelod:
There is currently a bug in the code that prints threshold pressures between all flow tubes if there is more than one equilibrium region. Thus many threshold pressures are added within one equilibrium region as well (probably from when individual and global were the only available options). This PR makes it such that THPRES is only printed for neighboring equilibrium regions (sorted for readability).

---

### Contributor checklist

- [x] :tada: This PR closes #400.
   - [x] Add conversion from finding combination of tubes connected at nodes to find combination of regions connected at nodes

### DIFF
--- a/src/flownet/parameters/_equilibration.py
+++ b/src/flownet/parameters/_equilibration.py
@@ -113,7 +113,22 @@ class Equilibration(Parameter):
             for connections_at_node in self._network.connection_at_nodes:
                 eqlnum_combinations.extend(list(combinations(connections_at_node, 2)))
 
-            eqlnum_combinations = list(set(eqlnum_combinations))
+            # Go from tube index to equilibrium region index (0-index to 1-index)
+            # remove duplicates, and only one way thpres needed
+            eqlnum_combinations = list(
+                {
+                    (
+                        self._eqlnum.loc[tube_a, "EQLNUM"],
+                        self._eqlnum.loc[tube_b, "EQLNUM"],
+                    )
+                    for _, [tube_a, tube_b] in enumerate(eqlnum_combinations)
+                    if self._eqlnum.loc[tube_b, "EQLNUM"]
+                    > self._eqlnum.loc[tube_a, "EQLNUM"]
+                }
+            )
+            # sort for readability
+            eqlnum_combinations.sort()
+
             eqlnum1 = list(list(zip(*eqlnum_combinations))[0])
             eqlnum2 = list(list(zip(*eqlnum_combinations))[1])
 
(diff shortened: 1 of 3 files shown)


src/flownet/templates/THPRES.jinja2
```diff
@@ -2,6 +2,6 @@
 -- FROM TO VALUE
 THPRES
 {% for i in range(eqlnum1|length) -%}
- {{ eqlnum1[i] + 1 }} {{ eqlnum2[i] + 1 }} 1* /
+ {{ eqlnum1[i]}} {{ eqlnum2[i]}} 1* /
```
wouterjdb, line 5: Why is this not needed anymore? Does this not fail for the individual equil?

src/flownet/parameters/_equilibration.py
```diff
@@ -113,7 +113,22 @@ def render_output(self) -> Dict:
...
+            # remove duplicates, and only one way thpres needed
+            eqlnum_combinations = list(
+                {
+                    (
+                        self._eqlnum.loc[tube_a, "EQLNUM"],
+                        self._eqlnum.loc[tube_b, "EQLNUM"],
+                    )
+                    for _, [tube_a, tube_b] in enumerate(eqlnum_combinations)
+                    if self._eqlnum.loc[tube_b, "EQLNUM"]
+                    > self._eqlnum.loc[tube_a, "EQLNUM"]
```
wouterjdb, line 126: Clever, this only adds THPRES it in one direction.

src/flownet/templates/THPRES.jinja2
olelod: The individual option will still go through this conversion
```
 eqlnum_combinations = list(
                {
                    (
                        self._eqlnum.loc[tube_a, "EQLNUM"],
                        self._eqlnum.loc[tube_b, "EQLNUM"],
                    )
                    for _, [tube_a, tube_b] in enumerate(eqlnum_combinations)
                    if self._eqlnum.loc[tube_b, "EQLNUM"]
                    > self._eqlnum.loc[tube_a, "EQLNUM"]
                }
            )
```
where you go from tube number to equilibrium region number (for individual this will be equal to +1)

wouterjdb (approved):
LGTM 🚀

